PR #2, opened 2024-07-12
Porting channel abstraction

jaehunkim:
Still in draft.
Porting an OOP code to non-OOP language is confusing 😵‍💫

https://www.notion.so/Implement-Channel-Abstraction-a0dad2aef762453783a08a83a3abcac6?pvs=4


src/channel/mod.rs
```diff
@@ -11,4 +18,101 @@ trait Channel {
...
+    }
+
+    fn begin_query_phase(&mut self) {
+        self.get_state_mut().is_query_phase = true;
+    }
+
+    /// Channel statistics related methods
+    fn get_state(&self) -> &ChannelStates;
+    /// XXX : dunno if this is the right approach
+    fn get_state_mut(&mut self) -> &mut ChannelStates;
```
jaehunkim: Expose mutational getter in trait, implement other default methods depending on it

src/channel/annotations.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn add_extra_annotation(&mut self, annotation: String) {
+        assert!(
+            self.extra_annotations_enabled,
+            "Cannot add extra annotation when extra annotations are disabled."
+        );
+        // XXX : why add prefix here, while add_annotation doesn't? XD
+        self.extra_annotations.push(format!("{}{}\n", self.annotation_prefix, annotation));
+    }
+
+    pub fn annotate_prover_to_verifier(&mut self, annotation: String, n_bytes: usize) {
```
rot256: Consider introducing an annotation type rather than having everything as String, ala.

```
enum Annotation {
    ProverToVerifier(n_bytes),
    VerifierToProver(..)
}
```

In general: sum types are one of the things which is much less painful in Rust than C++.

src/channel/mod.rs
rot256:
```
get_state(&self)
```

And

```
get_state_mut(&mut self)
```

Could be `AsRef` and `AsMut` implementations: https://doc.rust-lang.org/std/convert/trait.AsRef.html

src/channel/annotations.rs
rot256: This also prevents the need for e.g separate:

`annotate_prover_to_verifier`

And

`annotate_verifier_to_prover`

Methods

src/channel/mod.rs
```diff
@@ -11,4 +18,101 @@ trait Channel {
...
+    /// Only relevant for non-interactive channels. Changes the channel seed to a "safer" seed.
+    /// 
+    /// This function guarantees that randomness fetched from the channel after calling this function
+    /// and before sending data from the prover to the channel, is "safe": A malicious
+    /// prover will have to perform 2^security_bits operations for each attempt to randomize the fetched
+    /// randomness.
+    /// 
+    /// Increases the amount of work a malicious prover needs to perform, in order to fake a proof.
+    #[allow(unused_variables)]
+    fn apply_proof_of_work(&mut self, security_bits: usize) -> Option<()> {
```
rot256: I was thinking about how best to do this.

If the PoW is at the end of the transcript it might make sense for this to take ownership of the channel: it finishes the transcript.

In any case, it should return a `Result<(), SomeErrorType>` :)

rot256: Alternatively, since it is: 

"Only relevant for non-interactive channels"

You could introduce another trait:

```
trait FSChannel: Channel {
    fn apply_proof_of_work(&mut self, security_bits: usize) -> Option<()>;
}
```

That way we know at the type level which code is FS specific.

src/channel/verifier_channel.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn receive_commitment_hash(&mut self, annotation: &str) -> Vec<u8> {
+    }
+
+    pub fn recv_felt(&mut self, annotation: &str) -> F {
+    }
+
+    pub fn recv_data(&mut self, num_bytes: usize, annotation: &str) -> Vec<u8> {
+    }
+
+    pub fn dump_extra_annotations<W: Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
```
rot256: I would suggest separating the access to the struct member from the IO:

```
pub fn get_extra_annotations(&self) -> &[Annotation]
```

src/channel/mod.rs
jaehunkim: Yeah, I also think implementing an additional trait for non-interactive protocols is the best approach. 👍

src/channel/verifier_channel.rs
```diff
@@ -0,0 +1,78 @@
...
+        for annotation in &self.annotations.extra_annotations {
+            writeln!(writer, "{}", annotation)?;
+        }
+        Ok(())
+    }
+
+    pub fn annotate_extra_field_element(&mut self, field_element: F, annotation: &str) {
```
rot256: Again this could be enum variants.

In the stone-prover the analogous method is used twice:

```c++
// Annotate previous FRI layer for fri-proof-splitting.
--
147 | channel_->AnnotateExtraFieldElement(query_results_[j], ElementDecommitAnnotation(query_loc));
148 | // Compute and annotate the x-inv of query j for fri-proof-splitting.
149 | FieldElement x_inv = basis.GetFieldElementAt(query_index).Inverse();
150 | channel_->AnnotateExtraFieldElement(x_inv, "xInv for index " + std::to_string(query_index));
```

So I would introduce enum variants, one for each. The fact that they are "flattened" to a string in the stone-prover is not very readable.

src/channel/verifier_channel.rs
jaehunkim: This has been a habit I've been thinking about for a long- time.
I thought, 'If we're going to allow public access to member variables anyway, isn't it cleaner to access the variable directly rather than creating excessive getters and setters? Especially for small codebase?'
So I intentionally implemented it this way. 

However, I acknowledge that this implementation method can be confusing. 😓 
I'll use getters. 👍

src/channel/verifier_channel.rs
jaehunkim: Great! I strongly agree.
I was only thinking about how to adapt OOP impl. to Rust idioms, but I overlooked the fact that storing annotations as string continuously is a terrible design choice

src/channel/mod.rs
jaehunkim: Choose the first way, as there are already too many inheritances.
Note that there is no implementation for non-FS channels yet, though

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::hashutil::TempHashContainer;
+use ark_ff::Field;
+use std::marker::PhantomData;
+
+use rand_chacha::rand_core::RngCore;
+use rand_chacha::ChaCha20Rng;
+
+pub struct FSProverChannel<F: Field, H: TempHashContainer> {
+    _ph: PhantomData<(F, H)>,
+    prng: ChaCha20Rng,
```
rot256: We need to use the same PRNG as the stone-prover:

The transcript implementation(s) need to be 100% compatible with chose of the stone prover.

In this case the PRNGs used are covered in:

https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/randomness/prng.cc

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,112 @@
...
+        number % upper_bound
+    }
+
+    fn random_field(&mut self) -> Self::Field {
+        assert!(
+            !self.is_query_phase(),
+            "Prover can't receive randomness after query phase has begun."
+        );
+
+        let mut raw_bytes = [0u8; std::mem::size_of::<u64>()];
```
rot256: This is not enough bytes to sample a random field element :)

We also need to have exactly the same behavior as the stone-prover channel(s).

So we need to implement this:

https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/channel/noninteractive_prover_channel.cc#L74

src/channel/fs_prover_channel.rs
rot256: So Blake2s, Keccak256, Poseidon (work-in-progreess)

src/channel/fs_verifier_channel.rs
```diff
@@ -0,0 +1,106 @@
...
+        // TODO
+        Ok(felem)
+    }
+
+    fn recv_bytes(&mut self, n: usize) -> Result<Vec<u8>, anyhow::Error> {
+        assert!(
+            self.proof_read_index + n <= self.proof.len(),
+            "Proof too short."
+        );
+        let raw_bytes = self.proof[self.proof_read_index..self.proof_read_index + n].to_vec();
```
rot256: You need to add these to the transcript.

Additionally, you can turn the proof into an Iterator<Item = u8> or Reader which avoids you having to manually do the indexing:

https://doc.rust-lang.org/stable/std/io/struct.Cursor.html

src/channel/fs_verifier_channel.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.proof_read_index += n;
+        if !self.is_query_phase() {
+            // TODO : Mix seed with bytes
+        }
+        self.increment_byte_count(raw_bytes.len());
+        Ok(raw_bytes)
+    }
+
+    // send random number to prover
+    fn random_number(&mut self, upper_bound: u64) -> u64 {
```
rot256: This seems correctly implemented

src/channel/fs_verifier_channel.rs
```diff
@@ -0,0 +1,106 @@
...
+        assert!(
+            !self.is_query_phase(),
+            "Verifier can't send randomness after query phase has begun."
+        );
+
+        // TODO : change bytes size dynamically
```
rot256: I don't really follow this comment.

src/channel/fs_verifier_channel.rs
```diff
@@ -0,0 +1,106 @@
...
+        number % upper_bound
+    }
+
+    fn random_field(&mut self) -> Self::Field {
+        assert!(
+            !self.is_query_phase(),
+            "Verifier can't send randomness after query phase has begun."
+        );
+
+        let mut raw_bytes = [0u8; std::mem::size_of::<u64>()];
```
rot256: Same deal as the previous "random_field" implementation.

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
+
+// TODO : Find proper hash libaray
+// Digest? hashes
+pub trait TempHashContainer {
```
rot256: This one maybe:

https://docs.rs/digest/latest/digest/trait.Digest.html

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+    }
+}
+
+trait VerifierChannel: Channel {
+    type HashT: TempHashContainer;
+
+    fn recv_commit_hash(&mut self) -> Result<Self::HashT, anyhow::Error> {
+        let bytes = self.recv_bytes(Self::HashT::size())?;
+        let mut hash = Self::HashT::init_empty();
+        hash.update(&bytes);
```
rot256: This is not bound to the transcript hash.

In the stone-prover which method does this correspond to?

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+}
+
+trait ProverChannel: Channel {
+    type HashT: TempHashContainer;
+
+    fn send_felts(&mut self, felts: Vec<Self::Field>) -> Result<(), anyhow::Error>;
+
+    fn send_bytes(&mut self, bytes: Vec<u8>) -> Result<(), anyhow::Error>;
+
+    fn send_commit_hash(&mut self, hash: Self::HashT) -> Result<(), anyhow::Error> {
```
rot256: I think you should separate the "Output" type of a hash function from the instance (to which you can e.g. write).

See https://docs.rs/digest/latest/digest/trait.Digest.html for an example:

Digest has an associated type "Output" which you can turn into bytes.
I think this method should take an "Output".

src/hashutil/mod.rs
rot256: Although, it might not fit fully if you also want to capture:

https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/channel/noninteractive_prover_felt_channel.cc

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+    fn init_empty() -> Self;
```
rot256: If you still find this trait useful, this should probably be a "Default" implementation.

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    fn init_digest(data: &Vec<u8>) -> Self;
```
rot256: I would get rid of this, or have a default implementation:

```rust
fn digest(data: &[u8]) -> Self {
    let mut hash = Self::default();
    hash.update(data);
    hash
}
```

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    fn update(&mut self, data: &Vec<u8>);
```
rot256: This should take a slice: `&[u8]`

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    fn hash(&self) -> Vec<u8>;
```
rot256: This should return an "Output"

src/hashutil/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    fn size() -> usize;
```
rot256: This should be an associated constant on the "Output" type.

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,112 @@
...
+use rand_chacha::ChaCha20Rng;
+
+pub struct FSProverChannel<F: Field, H: TempHashContainer> {
+    _ph: PhantomData<(F, H)>,
+    prng: ChaCha20Rng,
+    proof: Vec<u8>,
+    states: ChannelStates,
+}
+
+impl<F: Field, H: TempHashContainer> AsMut<ChannelStates> for FSProverChannel<F, H> {
```
rot256: Do we need to expose `ChannelStates` at mut? 
Does it need to be changed outside the channel which "owns" it?

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+pub mod fs_prover_channel;
+
 use ark_ff::Field;
+use channel_states::ChannelStates;
+use std::convert::{AsMut, AsRef};
+use crate::hashutil::TempHashContainer;
 
 #[allow(dead_code)]
-trait Channel {
+trait Channel: AsRef<ChannelStates> + AsMut<ChannelStates> {
```
rot256: Does it need to be `AsMut`?

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+    /// Only relevant for non-interactive channels. Changes the channel seed to a "safer" seed.
+    ///
+    /// This function guarantees that randomness fetched from the channel after calling this function
+    /// and before sending data from the prover to the channel, is "safe": A malicious
+    /// prover will have to perform 2^security_bits operations for each attempt to randomize the fetched
+    /// randomness.
+    ///
+    /// Increases the amount of work a malicious prover needs to perform, in order to fake a proof.
+    #[allow(unused_variables)]
+    fn apply_proof_of_work(&mut self, security_bits: usize) -> Result<(), anyhow::Error> {
```
rot256: I would consider adding a trait which inherits from the channel with this method:


```rust
trait FSChannel: Channel {
    fn apply_proof_of_work(&mut self, security_bits: usize) -> Result<(), anyhow::Error>;
}
```

That way you don't need to have implementations what just error :)

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+    fn increment_byte_count(&mut self, n: usize) {
+        AsMut::<ChannelStates>::as_mut(self).byte_count += n;
+    }
+
+    fn increment_commitment_count(&mut self) {
+        AsMut::<ChannelStates>::as_mut(self).commitment_count += 1;
+    }
+
+    fn increment_hash_count(&mut self) {
+        AsMut::<ChannelStates>::as_mut(self).hash_count += 1;
```
rot256: Just do `self.states.increment_hash_count += 1` etc.

`AsMut` and `AsRef` exists to expose members of a struct to the other consumers.

src/channel/mod.rs
```diff
@@ -1,14 +1,89 @@
...
+
+    fn is_query_phase(&self) -> bool {
+        AsRef::<ChannelStates>::as_ref(self).is_query_phase
+    }
+
+    fn begin_query_phase(&mut self) {
+        AsMut::<ChannelStates>::as_mut(self).is_query_phase = true;
+    }
+
+    fn increment_byte_count(&mut self, n: usize) {
```
rot256: Does this need to be part of the trait?

Is the channel not responsible for updating its own state? 
Or does that need to happen externally?

src/channel/fs_prover_channel.rs
jaehunkim: This implementation was created because I wanted to enforce that the `ChannelState` must be included in the struct when implementing the `Channel` trait.

It seems that the problem becomes simpler if I keep the `Channel` trait and `ChannelStatus` trait separate.
No needs to expose `ChannelStates` at mut then.

src/channel/mod.rs
jaehunkim: `ChannelState` has no meaning on its own and only becomes meaningful when it's part of a `Channel`. 
So I tried putting the logic for changing `ChannelState` into the `Channel` trait.

It seems there's no real need to add dependencies between these two traits. 
I've removed all of them and put everything back in its original place.

src/channel/mod.rs
jaehunkim: Answered : https://github.com/zksecurity/pumice/pull/2#discussion_r1684585436

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
+use crate::randomness::hash_chain::HashChain;
+use ark_ff::BigInteger;
+use ark_ff::PrimeField;
+use std::collections::HashSet;
+use std::time::{SystemTime, UNIX_EPOCH};
+use std::vec::Vec;
+
+pub struct Prng {
```
rot256: Should this actually be a trait?

You could then implement `Prng` for `HashChain`.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        let hello_world = b"Hello World!";
+        let hash_ch_1 = HashChain::new_with_public_input(hello_world);
+        let hash_ch_2 = HashChain::default();
+
+        assert_ne!(
+            hash_ch_2.get_hash_chain_state(),
+            hash_ch_1.get_hash_chain_state()
+        );
+
+        let exp_hw_hash = [
```
rot256: Where does this constant come from?

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn uniform_int_vec(&mut self, min: u64, max: u64, n_elements: usize) -> Vec<u64> {
+        assert!(min <= max, "Invalid interval");
+        let mut return_vec = Vec::with_capacity(n_elements);
+        for _ in 0..n_elements {
+            return_vec.push(self.uniform_int(min, max));
+        }
+        return_vec
+    }
+
+    pub fn uniform_distinct_int_vec(&mut self, min: u64, max: u64, n_elements: usize) -> Vec<u64> {
```
rot256: If you turn this into a trait, you can implement this generically for every implementation of the trait or provide a default implementation on the trait.

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+            let value = self.uniform_int(min, max);
+            if current_set.insert(value) {
+                return_vec.push(value);
+            }
+        }
+        return_vec
+    }
+
+    // XXX : dumb implementation. should reduce calls to uniform_int_vec
+    pub fn uniform_bool_vec(&mut self, n_elements: usize) -> Vec<bool> {
```
rot256: Same deal

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+        // or we can use F::MODULUS_MINUS_ONE_DIV_TWO * 2
+
+        for _ in 0..n_elements {
+            let value = self.uniform_bigint(min, max);
+            return_vec.push(F::from_bigint(value).unwrap());
+        }
+        return_vec
+    }
+
+    pub fn seed_from_system_time() -> [u8; 8] {
```
rot256: When is this used?

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+use ark_ff::PrimeField;
+use std::collections::HashSet;
+use std::time::{SystemTime, UNIX_EPOCH};
+use std::vec::Vec;
+
+pub struct Prng {
+    hash_chain: HashChain,
+}
+
+// TODO : Implement Rng trait
```
rot256: This seems like a good idea, in particular `fill_bytes`

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        // Serialized counter is in big-endian format.
+        bytes_with_counter[KECCAK256_DIGEST_NUM_BYTES * 2 - 8..]
+            .copy_from_slice(&self.counter.to_be_bytes());
+
+        let mut hasher = Keccak256::new();
+        hasher.update(bytes_with_counter);
+        let result = hasher.finalize();
+        let mut hash_bytes = [0u8; KECCAK256_DIGEST_NUM_BYTES];
+        hash_bytes.copy_from_slice(&result);
+        hash_bytes
```
rot256: Imo, makes sense to update the `counter` here rather than make it the callers responsibility to remember.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    fn next_hash(&self) -> [u8; KECCAK256_DIGEST_NUM_BYTES] {
+        // TODO: below code is not efficient, but it works for now
+        let mut bytes_with_counter = [0u8; KECCAK256_DIGEST_NUM_BYTES * 2];
+        bytes_with_counter[..KECCAK256_DIGEST_NUM_BYTES].copy_from_slice(&self.digest);
+
+        // Copy the counter's serialized 64bit onto the MSB end of the buffer (stone-prover PR #875 decision).
+        // Serialized counter is in big-endian format.
+        bytes_with_counter[KECCAK256_DIGEST_NUM_BYTES * 2 - 8..]
```
rot256: Just to check: there is a bunch of zeroes in the middle?

```
Digest || 0x0^(256 - 8) || BigEndian(Counter)
```

src/randomness/hash_chain.rs
rot256: Considering how it is used, maybe have it set:

```
self.spare_bytes.copy_from_slice(&result)
self.num_spare_bytes = KECCAK256_DIGEST_NUM_BYTES
```

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        hash_bytes.copy_from_slice(&result);
+        Self {
+            digest: hash_bytes,
+            spare_bytes: [0u8; KECCAK256_DIGEST_NUM_BYTES * 2],
+            num_spare_bytes: 0,
+            counter: 0,
+        }
+    }
+
+    pub fn reseed(&mut self, public_input: &[u8]) {
```
rot256: I would consider removing this and just using the constructor above to simplify the interface.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        let big_int = U256::from_be_bytes(self.digest).add(U256::from(seed_increment));
+
+        // Serialize the incremented big_int back into the mixed_bytes
+        mixed_bytes[..KECCAK256_DIGEST_NUM_BYTES].copy_from_slice(&big_int.to_be_bytes());
+
+        // Copy the raw_bytes into the mixed_bytes
+        mixed_bytes[KECCAK256_DIGEST_NUM_BYTES..].copy_from_slice(raw_bytes);
+
+        // Hash the mixed_bytes to update the digest
+        let mut hasher = Keccak256::new();
```
rot256: We are doing this a lot:

```
let mut hasher = Keccak256::new();
hasher.update(&mixed_bytes);
let result = hasher.finalize();
self.digest.copy_from_slice(&result);
```

Maybe define:

```
#[inline(always)]
fn keccak256(bytes: &[u8]) -> [u8; KECCAK256_DIGEST_NUM_BYTES] {
  let mut hasher = Keccak256::new();
  hasher.update(&bytes);
  hasher.finalize().into()
}
```

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        hasher.update(&mixed_bytes);
+        let result = hasher.finalize();
+        self.digest.copy_from_slice(&result);
+
+        self.num_spare_bytes = 0;
+        self.counter = 0;
+    }
+
+    pub fn get_hash_chain_state(&self) -> &[u8; KECCAK256_DIGEST_NUM_BYTES] {
```
rot256: Do we need to expose this? Is it only used in tests?

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        let mut mixed_bytes = vec![0u8; KECCAK256_DIGEST_NUM_BYTES + raw_bytes.len()];
+
+        // Deserialize the current digest into a u64 array
+        let big_int = U256::from_be_bytes(self.digest).add(U256::from(seed_increment));
+
+        // Serialize the incremented big_int back into the mixed_bytes
+        mixed_bytes[..KECCAK256_DIGEST_NUM_BYTES].copy_from_slice(&big_int.to_be_bytes());
+
+        // Copy the raw_bytes into the mixed_bytes
+        mixed_bytes[KECCAK256_DIGEST_NUM_BYTES..].copy_from_slice(raw_bytes);
```
rot256: No need for a copy:

```rust
let mut hasher = Keccak256::new();
hasher.update(&big_int.to_be_bytes());
hasher.update(raw_bytes);
```

src/randomness/hash_chain.rs
rot256: Alternatively if you want a general macro (capable of handling any number of inputs), you can use:


```rust
macro_rules! keccak256 {
    ( $($input:expr),* ) => {{
        use keccak256::Digest;
        let mut hsh = Keccak256::new();
        $(
            hsh.update($input);
        )*
        let out: [u8; KECCAK256_DIGEST_NUM_BYTES] = hsh.finalize().into();
        out
    }};
}
```

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub fn get_hash_chain_state(&self) -> &[u8; KECCAK256_DIGEST_NUM_BYTES] {
+        &self.digest
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const RANDOM_BYTES_1ST_KECCAK256_TEST: [u8; 8] =
```
rot256: Where do these test vectors come from btw?

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.digest = hash_bytes;
+        self.num_spare_bytes = 0;
+        self.counter = 0;
+    }
+
+    pub fn random_bytes(&mut self, random_bytes_out: &mut [u8]) {
+        let num_bytes = random_bytes_out.len();
+        let num_full_blocks = num_bytes / KECCAK256_DIGEST_NUM_BYTES;
+
+        for offset in
```
rot256: There is a lot of indexing going on there, kinda hard to follow, maybe using https://doc.rust-lang.org/std/slice/struct.ChunksExact.html is an option.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+
+            self.spare_bytes.copy_within(num_tail_bytes.., 0);
+        } else {
+            self.fill_random_bytes(
+                &mut random_bytes_out[num_full_blocks * KECCAK256_DIGEST_NUM_BYTES..num_bytes],
+            );
+        }
+    }
+
+    fn fill_random_bytes(&mut self, out: &mut [u8]) {
```
rot256: There is a lot of complexity introduced by the "spilling over" behavior and trying to implement this efficiently:

One option is to see if you can use https://docs.rs/rand_core/latest/rand_core/block/struct.BlockRng.html

Alternatively, do a more naive implementation using a byte-by-byte loop: this code is not performance critical, all in all it needs to hash a few hundreds of KB and generate maybe a KB of output.

So readability is more important than performance here.

src/randomness/hash_chain.rs
rot256: Nitpick: maybe call it something different than `big_int` which is its type.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        Self {
+            digest: [0u8; KECCAK256_DIGEST_NUM_BYTES],
+            spare_bytes: [0u8; KECCAK256_DIGEST_NUM_BYTES * 2],
+            num_spare_bytes: 0,
+            counter: 0,
+        }
+    }
+}
+
+impl HashChain {
```
rot256: Should this be generic over a `Digest`?

I think we also need `Blake2s`?

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub fn update_hash_chain(&mut self, raw_bytes: &[u8]) {
+        let seed_increment: u64 = 0;
+        self.mix_seed_with_bytes(raw_bytes, seed_increment);
+    }
+
+    pub fn mix_seed_with_bytes(&mut self, raw_bytes: &[u8], seed_increment: u64) {
+        let mut mixed_bytes = vec![0u8; KECCAK256_DIGEST_NUM_BYTES + raw_bytes.len()];
+
+        // Deserialize the current digest into a u64 array
+        let big_int = U256::from_be_bytes(self.digest).add(U256::from(seed_increment));
```
rot256: I think this:

https://docs.rs/ark-ff/latest/ark_ff/biginteger/trait.BigInteger.html#tymethod.add_with_carry

Might be enough and we might be able to avoid the dependency on ethnum.

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    // template <typename OtherHashT>
+    // OtherHashT RandomHash() {
+    //   return OtherHashT::InitDigestTo(RandomByteVector(OtherHashT::kDigestNumBytes));
+    // }
+
+    // pub fn random_other_hash
+
+    pub fn mix_seed_with_bytes(&mut self, raw_bytes: &[u8]) {
+        let seed_increment: u64 = 1;
```
rot256: Add:

```
const INCREMENT_SEED: u64 = 1
```

To the top.

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,243 @@
...
+        hash_bytes.copy_from_slice(&result);
+        hash_bytes
+    }
+
+    pub fn update_hash_chain(&mut self, raw_bytes: &[u8]) {
+        let seed_increment: u64 = 0;
+        self.mix_seed_with_bytes(raw_bytes, seed_increment);
+    }
+
+    pub fn mix_seed_with_bytes(&mut self, raw_bytes: &[u8], seed_increment: u64) {
```
rot256: Since it overwrites the entire state of the object, consider if it should actually take ownership of the object and return a new instance rather than have side-effects, e.g.


```rust
pub fn mix(self, bytes: &[u8], sep: u64) -> Self {
    let mut hasher = Keccak256::new();
    hasher.update(
        U256::from_be_bytes(self.digest)
            .add(U256::from(sep))
            .to_be_bytes(),
    );
    hasher.update(bytes);
    Self {
        digest: hasher.finalize().into(),
        ..Default::default()
    }
}
```

Or with the macro:

```rust
pub fn mix(self, bytes: &[u8], sep: u64) -> Self {
    Self {
        digest: keccak256!(
          U256::from_be_bytes(self.digest)
              .add(U256::from(sep))
              .to_be_bytes(),
           bytes
        ),
        ..Default::default()
    }
}
```

src/randomness/hash_chain.rs
rot256: If all we need is an addition maybe we should just implement that ourselves:

Adding the low 64-bits and carrying forward the overflow should only be a couple lines.

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+                }
+            }
+
+            // then make it to BigInteger
+            random_value = B::from_bits_le(&bits);
+            if random_value <= range {
+                break;
+            }
+        }
+        random_value.add_with_carry(&min);
```
rot256: Is this correct? The `range` is already `max` so addming `min` takes it out of range no?

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn uniform_bigint<B: BigInteger>(&mut self, min: B, max: B) -> B {
+        assert!(min <= max, "Invalid interval");
+        let mut range = max.clone();
+        range.sub_with_borrow(&min);
+        let mut random_value: B;
+
+        let num_bits = range.num_bits() as usize;
+
+        loop {
+            let mut bytes = vec![0u8; B::NUM_LIMBS * 8]; // B::NUM_LIMBS * sizeof(u64)
```
rot256: `std::mem::size_of::<u64>()`

src/randomness/prng.rs
```diff
@@ -0,0 +1,297 @@
...
+        return_vec
+    }
+
+    // XXX : dumb implementation. should reduce calls to uniform_int_vec
+    pub fn uniform_bool_vec(&mut self, n_elements: usize) -> Vec<bool> {
+        let bits = self.uniform_int_vec(0, 1, n_elements);
+        bits.into_iter().map(|bit| bit != 0).collect()
+    }
+
+    pub fn uniform_bigint<B: BigInteger>(&mut self, min: B, max: B) -> B {
```
rot256: This corresponds to https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/algebra/fields/long_field_element.cc#L61 right?

src/randomness/prng.rs
rot256: Another nitpick, there is already:

https://doc.rust-lang.org/std/ops/struct.Range.html

So you could make the function take a range, which allows for semantic sugar like:

```
channel.uniform_distinct_int_vec(0..end, num);
```

src/randomness/hash_chain.rs
jaehunkim: From [testcase](https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/randomness/hash_chain_test.cc#L115) of `stone-prover`

src/randomness/hash_chain.rs
jaehunkim: Right. I followed [this](https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/randomness/hash_chain.cc#L126)

src/randomness/prng.rs
jaehunkim: This corresponds to https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/randomness/prng.h#L147

src/randomness/hash_chain.rs
jaehunkim: That's correct. However, until we run Fri, we only need an implementation for keccak256, so I worked on it this way for now. 😅

src/randomness/hash_chain.rs
jaehunkim: From stone-prover [testcase](https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/randomness/hash_chain_test.cc#L31)

src/randomness/hash_chain.rs
jaehunkim: It seems better to include incrementing the counter in this function. 
However, I think it's a better design in terms of separation of concerns for next_hash to only perform the creation of a new hash based on the current state, while the processing related to spare bytes is handled in fill_random_bytes.

src/randomness/hash_chain.rs
jaehunkim: It's for tests and verifying PoW.

src/randomness/hash_chain.rs
jaehunkim: Used it!

src/randomness/hash_chain.rs
jaehunkim: I used ethnum U256 because there's no from_be_bytes method in ark_ff::BigInteger. There's only from_bits_be 😓. I thought all the process of changing bytes into bits and then... is kind of redundant.
Implementing ADC ourselves seems to be the best choice out of all options. 👍

src/channel/fs_verifier_channel.rs
jaehunkim: nvm, It's a mistake

src/channel/mod.rs
jaehunkim: The implementation has moved, but this was following here : 
https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/channel/verifier_channel.h#L77

src/channel/mod.rs
jaehunkim: Got it !

src/hashutil/mod.rs
jaehunkim: Still, I'm not sure if we need a wrapper trait for hash function. Will check this again when I implement [noninteractive_prover_felt_channel.cc](https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/channel/noninteractive_prover_felt_channel.cc)

src/randomness/prng.rs
jaehunkim: There's `range.sub_with_borrow(&min);`

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,252 @@
...
+            random_biguint = BigUint::from_bytes_be(&raw_bytes);
+            if random_biguint < max_divisible {
+                random_biguint = random_biguint.modpow(&BigUint::from(1u64), &mod_felt252);
+                break;
+            }
+        }
+
+        let field_element: F = Self::Field::from_bigint(
+            <Self::Field as PrimeField>::BigInt::try_from(random_biguint).unwrap(),
+        )
```
jaehunkim: @rot256 
I had to use BigUint to divide some large integers (smaller than max_divisible) by mod_felt252. 
The current code is messy and lacks generality. 

I've been trying to find a way to calculate max_divisible once and assign it to a static constant, but this has been time-consuming. For now, I've left the implementation as is.

jaehunkim: Also, I cannot use `Self::Field::from_be_bytes_mod_order()`, output differs 😓

jaehunkim:
```rust
impl<F: PrimeField, D: Digest, P: Prng> FSProverChannel<F, D, P> {
    fn modulus() -> &'static BigUint {
        static MODULUS: OnceLock<BigUint> = OnceLock::new();
        MODULUS.get_or_init(|| F::MODULUS.into())
    }

    fn max_divislble() -> &'static BigUint {
        static MAX_VALUE: OnceLock<BigUint> = OnceLock::new();
        MAX_VALUE.get_or_init(|| {
            let modulus = F::MODULUS.into();
            let size: usize = ((F::MODULUS_BIT_SIZE + 7) / 8) as usize;
            let max = BigUint::from_bytes_be(&vec![0xff; size]);
            let quotient = max.div(&modulus);
            quotient * modulus
        })
    }
}
```

Removed non-generic constant.

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,252 @@
...
+        let mod_felt252 = BigUint::parse_bytes(
+            b"800000000000000000000000000000000000000000040000000000000000001",
+            16,
+        )
+        .unwrap();
+
+        let mut raw_bytes: Vec<u8>;
+        let mut random_biguint: BigUint;
+        loop {
+            raw_bytes = self.draw_bytes(((Self::Field::MODULUS_BIT_SIZE + 7) / 8) as usize);
```
rot256: https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,252 @@
...
+            random_biguint = BigUint::from_bytes_be(&raw_bytes);
+            if random_biguint < max_divisible {
+                random_biguint = random_biguint.modpow(&BigUint::from(1u64), &mod_felt252);
```
rot256: https://docs.rs/num-integer/0.1.46/num_integer/trait.Integer.html#tymethod.div_rem

Maybe cleaner.

src/channel/fs_prover_channel.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    fn draw_felem(&mut self) -> Self::Field {
+        assert!(
+            !self.states.is_query_phase(),
+            "Prover can't receive randomness after query phase has begun."
+        );
+
+        // from stone-prover/src/starkware/algebra/fields/big_prime_constants.h
+        //   static constexpr ValueType kMaxDivisible = 0xf8000000000000000000000000000000000000000007c000000000000000001f_Z;
+        let max_divisible = BigUint::parse_bytes(
```
rot256: Alternatively, you could use something cheaper like:

https://docs.rs/num-bigint/0.4.3/num_bigint/struct.BigUint.html#method.from_slice

src/channel/fs_prover_channel.rs
rot256: Ah damn, curious, any idea why the output from `Self::Field::from_be_bytes_mod_order()` differs?

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+                    {
+                        break;
+                    }
+                }
+            }));
+        }
+
+        for handle in handles {
+            handle.join().unwrap();
+        }
```
rot256: Consider using scoped threads:

https://doc.rust-lang.org/std/thread/fn.scope.html

rot256: This should also allow you to avoid using `Arc` as scoped threads can borrow things which are not `'static`: Rust knows that e.g. `next_chunk_to_search` will live longer than the threads, since they *must* be joined.

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        let nonce_as_u64 = lowest_nonce_found.load(Ordering::Relaxed);
+        assert!(nonce_as_u64 != u64::MAX, "No nonce was found.");
+        nonce_as_u64.to_be_bytes().to_vec()
+    }
+
+    fn init_hash(&self, seed: &[u8], work_bits: usize) -> D {
+        let mut hasher = D::new();
+        hasher.update(&[0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xed]);
```
rot256: Move this magic into a constant :)

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+        hasher.update(seed);
+        hasher.update(&[work_bits as u8]);
+        hasher
+    }
+}
+
+fn search_chunk<D: Digest>(
+    nonce_start: u64,
+    chunk_size: u64,
+    thread_bytes: &mut [u8],
```
rot256: For the sake of readability I would take this by copy: take a &[u8] then clone it.

That way it does not look like it "returns" a result in `thread_bytes`

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+    work_limit: u64,
+) -> Option<u64> {
+    let thread_len = thread_bytes.len();
+    for nonce in nonce_start..nonce_start + chunk_size {
+        thread_bytes[thread_len - 8..].copy_from_slice(&nonce.to_be_bytes());
+
+        let hash = D::new().chain_update(&thread_bytes).finalize();
+        let digest_word = u64::from_be_bytes(hash.as_slice()[..8].try_into().unwrap());
```
rot256: Maybe `size_of::<u64>()`

https://doc.rust-lang.org/std/mem/fn.size_of.html

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+        let init_hash = self.init_hash(seed, work_bits);
+        let output_size = <D as OutputSizeUser>::output_size();
+
+        let mut bytes = vec![0u8; output_size + 8];
+        bytes[..output_size].copy_from_slice(&init_hash.finalize());
+        bytes[output_size..].copy_from_slice(nonce_bytes);
+
+        let work_limit = 1u64 << (64 - work_bits);
+
+        let hash = D::new().chain_update(&bytes).finalize();
```
rot256: If you separate this into a function:

```
#[inline]
fn pow_check(bytes: &[u8]) -> u64 {
   let hash = D::new().chain_update(&bytes).finalize();
   let digest_word = u64::from_be_bytes(hash.as_slice()[..8].try_into().unwrap());
   digest_word
}
```

Then you can reuse it in the solver :)

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+        let digest_word = u64::from_be_bytes(hash.as_slice()[..8].try_into().unwrap());
+
+        digest_word < work_limit
+    }
+
+    fn init_hash(&self, seed: &[u8], work_bits: usize) -> D {
+        let mut hasher = D::new();
+        hasher.update(&[0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xed]);
```
rot256: Move this magic into a constant :)

src/channel/pow.rs
```diff
@@ -0,0 +1,259 @@
...
+            _hash: std::marker::PhantomData,
+        }
+    }
+
+    // TODO : Implement task manager
+    pub fn prove(
+        &self,
+        seed: &[u8],
+        work_bits: usize,
+        try_thread_count: usize,
```
rot256: Can we use:

https://doc.rust-lang.org/std/thread/fn.available_parallelism.html

Instead?

src/randomness/hash_chain.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+impl HashChain {
+    pub fn new_with_public_input(public_input: &[u8]) -> Self {
+        let result = keccak256!(public_input);
+        let mut hash_bytes = [0u8; KECCAK256_DIGEST_NUM_BYTES];
+        hash_bytes.copy_from_slice(&result);
+        Self {
+            digest: hash_bytes,
```
rot256: Maybe:

```
Self {
   digest: keccak256!(public_input).try_into().unwrap(),
   ..Default::default()
}
```

src/channel/fs_prover_channel.rs
rot256: Code above looks nice :)

src/channel/tests.rs
```diff
@@ -0,0 +1,252 @@
...
+use crate::{
+    randomness::prng::{Prng, PrngKeccak256, PrngOnlyForTest},
+    Felt252,
+};
+use ark_ff::PrimeField;
+use blake2::Blake2s256;
+use hex_literal::hex;
+use sha3::digest::generic_array::GenericArray;
+use sha3::digest::OutputSizeUser;
+use sha3::Sha3_256;
```
rot256: Wait a sec... they use both Sha3 and the pre-standard Keccak256?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
rot256: I don't know about adding .vscode to the repo.

rot256 (approved):
Overall it looks good, I think in the interest of getting this in, I think we should merge this rather large PR and then maybe iterate on the code: e.g. make it generic in a subsequent PR.

rot256:
Is it possible to update the commit messages to adhere to the convention and maybe squash commits like:

- Applied code-reviews
- Applied code-reviews2

src/channel/pow.rs
jaehunkim: Even when using scoped threads, shouldn't the part that updates `next_chunk_to_search` be thread-safe? 

While using scoped threads has the advantage of not needing to copy `thread_bytes` or some integer values, it seems that atomic operations are still necessary. 👀

jaehunkim:
@rot256 

I should have handled rebasing well from the beginning and applied commit conventions, but I failed to do so. 

I tried to rebase and organize this giant PR, but it seems too complicated.
For this PR, I'll do a squash merge to avoid cluttering the master branch.
From now on, I'll follow conventions and do appropriate squashing.

src/channel/tests.rs
jaehunkim: https://github.com/zksecurity/pumice/pull/5

It was only to indicate the type of arguments for `recv_commit_hash` or `send_commit_hash`. I eliminated it.

PR #3, opened 2024-07-18
Implemented and Tested Starkware Poseidon

rot256:
Implements the Starkware-flavor Poseidon hash function.

Also restructures the project into sub-crates.


rot256:
Btw, performance seems *pretty* good, almost twice as fast as https://docs.rs/starknet-crypto/latest/starknet_crypto/ on the same hardware:

```
Poseidon3::pair         time:   [2.8650 µs 2.8725 µs 2.8805 µs]
                        change: [-0.0362% +0.1902% +0.4267%] (p = 0.09 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
```

poseidon/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        //  ( 1,-1, 1),
+        //  ( 1, 1, -2)
+        // )
+        // see:
+        // https://github.com/starkware-industries/poseidon/blob/main/poseidon3.txt
+        let t0 = st[0] + st[1];
+        let t1 = t0 + st[2];
+        st[0] = t1 + st[0].double();
+        st[1] = t1 - st[1].double();
+        st[2] = t0 - st[2].double();
```
jaehunkim, line 44: Compare to https://github.com/lambdaclass/lambdaworks/blob/main/crypto/src/hash/poseidon/starknet/parameters.rs#L25
You replaced one addition operation to one allocation. :+1:

If replacing one addition to one allocation gives better result in lamdaworks codebase, it might be good to submit a PR to the lambdaworks implementation. (Is there a conflict of interest?)

poseidon/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    fn round_partial<'a>(st: &mut [Felt252; WIDTH], key: &mut impl Iterator<Item = &'a Felt252>) {
```
jaehunkim, line 47: ~Adding inline suggestion to `round_partial` and `round_full` gives me a slightly better(about 1%) performance~
nah, it's not reproducible, it seems to have been a momentary anomalous result.

poseidon/build.rs
```diff
@@ -0,0 +1,349 @@
...
+            .then(eq)
+            .then(mxt.clone())
+            .map(|(_, m)| Options::RoundKeys(m)),
+    ));
+
+    opt.padded().repeated().then(end()).map(|(v, _)| v)
+}
+
+#[allow(clippy::single_char_add_str)]
+fn main() {
```
jaehunkim: Is there any benefit to including builder code rather than pre-built code?
Or it's just a matter of choice? I'm just curios 😅

jaehunkim (commented):
Unfortunately, three days ago, starknet-rs changed to use poseidon from type-rs, or more precisely, the implementation from lambdaworks. 

When analyzing the overall logic, our implementation seems more(or at least similarly) efficient, but the performance is slightly lower. This is probably due to using a different Field crate.

LGTM!

poseidon/build.rs
rot256: I went back/forth on this: we could parse them once and include the compressed round keys.

- On one hand it would simplify things: just have an array with the processed round keys.
- On the other hand with would make it less clear where these constants come from, if, e.g. I had a separate python script producing them and then simply added these as constants.

It might be over engineered: roughly half the code is doing this.

rot256: Your right, I decided that it was simpler to just add the processed output.

PR #4, opened 2024-07-26
Implement Commitment Scheme Methods

varunthakore:
This PR implements commitment scheme methods corresponding to the following library of Stone.
https://github.com/starkware-libs/stone-prover/tree/main/src/starkware/commitment_scheme


merkle/src/hash.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub struct Keccak256Hasher<F: PrimeField> {
+    _ph: PhantomData<F>,
+}
+
+impl<F: PrimeField> Hasher<F> for Keccak256Hasher<F> {
+    type Output = [u8; 32];
+
+    fn leaf(input: &[F]) -> Self::Output {
+        let input_bytes: Vec<u8> = input
```
rot256: Maybe:

```
let mut hasher = Keccak256::new();
input.for_each(|f| hasher.update(f.into_bigint().to_bytes_be()));
hasher.finalize().into()
```

merkle/src/hash.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl Hasher<Felt252> for Poseidon3<Felt252> {
+    type Output = Felt252;
+
+    fn leaf(input: &[Felt252]) -> Self::Output {
+        Poseidon3::hash(input)
+    }
+
+    fn node(input: &[Self::Output]) -> Self::Output {
+        Poseidon3::hash(input)
```
rot256: Did you test this?

Asking, because they also have a distinct `Poseidon3::pair` (hashing two elements), which is not the same as `Poseidon3::hash` of two elements...

merkle/src/hash.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // compress a list of leaves into a single leaf
+    fn leaf(input: &[F]) -> Self::Output;
+}
+
+pub struct Blake2s256Hasher<F: PrimeField> {
+    _ph: PhantomData<F>,
+}
+
+impl<F: PrimeField> Hasher<F> for Blake2s256Hasher<F> {
```
rot256: I wonder if he can have a generic implementation for every `Digest`.

merkle/src/lib.rs
```diff
@@ -46,21 +74,68 @@ impl<F: Field, H: Hasher<F>> MerkleTree<F, H> {
...
-    #[allow(clippy::result_unit_err)]
     pub fn verify_decommitment(
         &self,
         comm: H::Output,
         to_verify: &[(usize, H::Output)],
-        siblings: impl Iterator<Item = H::Output>,
-    ) -> Result<(), ()> {
-        if self.root_decommitment(to_verify, siblings).ok_or(())? == comm {
-            Ok(())
+    ) -> Result<bool, MerkleError> {
```
rot256: This code assumes that you have all the nodes in the tree: the verifier does not.

merkle/src/lib.rs
```diff
@@ -46,21 +74,68 @@ impl<F: Field, H: Hasher<F>> MerkleTree<F, H> {
...
+        assert!(!to_verify.is_empty(), "Empty verify queries");
+
+        let queries_idx: Vec<usize> = to_verify.iter().map(|&(x, _)| x).collect();
+        let siblings = self
```
rot256: The verifier needs to be able to read these as a stream of siblings out of a proof transcript.

merkle/src/lib.rs
```diff
@@ -46,21 +74,68 @@ impl<F: Field, H: Hasher<F>> MerkleTree<F, H> {
...
+    /// Generates siblings from input query indices.
+    ///
+    /// # Arguments
+    ///
+    /// - `queries_idx`: an array of [position] to query
+    ///
+    /// # Returns
+    ///
+    /// - `Some(sib)`: vector of sibling hashes
+    fn generate_decommitment(&self, queries_idx: &[usize]) -> Option<Vec<H::Output>> {
```
rot256: I think it might be useful to separate the Merkle tree verifier logic, from the Merkle tree prover logic: 

The verifier will basically only have the root and the number of expected levels. His only operation is verifying (batch) openings.

The prover will have all the nodes in the tree and be able to generate decommitments.

rot256 (commented):
The code looks really nice!

But I think we need to separate the Merkle tree prover / verifier logic.

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn leaf(input: &[F]) -> Self::Output {
+        let mut hasher = Blake2s256::new();
+        input
+            .iter()
+            .for_each(|f| hasher.update(f.into_bigint().to_bytes_be()));
+        hasher.finalize()
+    }
+
+    fn node(input: &[Self::Output]) -> Self::Output {
+        let mut hasher = Blake2s256::new();
```
mellowcroc, line 52: assert that input len is 2?

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,84 @@
...
+    fn leaf(input: &[F]) -> Self::Output {
+        let mut hasher = Keccak256::new();
+        input
+            .iter()
+            .for_each(|f| hasher.update(f.into_bigint().to_bytes_be()));
+        hasher.finalize()
+    }
+
+    fn node(input: &[Self::Output]) -> Self::Output {
+        let mut hasher = Keccak256::new();
```
mellowcroc: ditto

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+
+    use channel::fs_verifier_channel::FSVerifierChannel;
+    use channel::{fs_prover_channel::FSProverChannel, ProverChannel};
+    use randomness::keccak256::PrngKeccak256;
+    use randomness::Prng;
+    use sha3::Sha3_256;
+
+    fn hex_to_generic_array(hex_str: &str) -> GenericArray<u8, U32> {
+        let mut hex_str = String::from(hex_str);
+        let padding_length = 64_i32.saturating_sub(hex_str.len() as i32);
```
mellowcroc, line 262: I think we only need one of the two paddings (64 or 32) here

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+        H: Hasher<F, Output = GenericArray<u8, U32>>,
+    {
+        let height = data0.len().ilog2() as usize;
+        let mut tree = MerkleTree::<F, H>::new(data0.len());
+
+        MerkleTree::add_data(&mut tree, data0, 0);
+        let root0 = tree.get_root(0);
+
+        MerkleTree::add_data(&mut tree, data1, 0);
+        let root1 = tree.get_root(height);
```
mellowcroc: should be ok to use `0` instead of `height` since `add_data` computes all the nodes

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+            hex_to_generic_array(
+                "6dc48ad654bc4a3e3c8f3270a987b2782af1707e78e1512018f16fdee124bdbd",
+            ),
+        ];
+        let root_exp = hex_to_generic_array(
+            "b823ac891cee85512521528e1d61cdcad829080392b63f393e46adc213862af9",
+        );
+        test_compute_root::<Felt252, Blake2s256Hasher<Felt252>>(&input, root_exp);
+
+        let input = [
```
mellowcroc: this is a duplicate of the above test case

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+            hex_to_generic_array(
+                "1c16cd6af66980b3f38f66a213a82b0c5d3480296d6315d12ea3862e72e29a8c",
+            ),
+        ];
+        let root_exp = hex_to_generic_array(
+            "57989eb39d929071d2f3a4400067d00e4d77671d9ed682a8cdd2d4e11d3f5aea",
+        );
+        test_compute_root::<Felt252, Blake2s256Hasher<Felt252>>(&input, root_exp);
+
+        let input = [
```
mellowcroc: ditto

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
```
mellowcroc, line 506: ditto

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+            hex_to_generic_array(
+                "6dc48ad654bc4a3e3c8f3270a987b2782af1707e78e1512018f16fdee124bdbd",
+            ),
+        ];
+        let root_exp = hex_to_generic_array(
+            "b823ac891cee85512521528e1d61cdcad829080392b63f393e46adc213862af9",
+        );
+        test_verify_true::<Felt252, Blake2s256Hasher<Felt252>>(&input, root_exp);
+
+        let input = [
```
mellowcroc: this is a duplicate of the above test case

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
+            hex_to_generic_array(
+                "1c16cd6af66980b3f38f66a213a82b0c5d3480296d6315d12ea3862e72e29a8c",
+            ),
+        ];
+        let root_exp = hex_to_generic_array(
+            "57989eb39d929071d2f3a4400067d00e4d77671d9ed682a8cdd2d4e11d3f5aea",
+        );
+        test_verify_true::<Felt252, Blake2s256Hasher<Felt252>>(&input, root_exp);
+
+        let input = [
```
mellowcroc: ditto

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2004 @@
...
```
mellowcroc: ditto

commitment_scheme/src/merkle/merkle_commitment_scheme.rs
```diff
@@ -0,0 +1,135 @@
...
+            segment_index * self.segment_length_in_elements(),
+        );
+    }
+
+    fn commit(&mut self) {
+        let height = self.tree.data_length.ilog2() as usize;
+        let comm = self
+            .tree
+            .get_root(height - self.segment_length_in_elements().ilog2() as usize);
+        let _ = self.channel.send_commit_hash(comm);
```
mellowcroc: call `unwrap()` here?

mellowcroc (commented):
Code looks great! 👍

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2014 @@
...
+
+        assert!(root0 != root1);
+    }
+
+    #[test]
+    fn test_diff_tree_diff_root_with() {
+        // test with Blake2s256
+        let data0 = [
+            hex_to_generic_array(
+                "837cd6db8f93d5ddfeaaba5509551c69c760c099e97969cfa04047ae65c33c8c",
```
mellowcroc: Can we reuse these constants across different test cases? (i.e. instantiate them once and reuse them?)

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,2014 @@
...
+    /// Retrieves the root of the tree assuming nodes at some depth are known.
+    ///
+    /// # Arguments
+    ///
+    /// - `min_depth_assumed_correct`: minimal depth (distance from the root) assumed to be correct.
+    ///
+    /// # Returns
+    ///
+    /// Returns the root of the tree.
+    pub fn get_root(&mut self, min_depth_assumed_correct: usize) -> H::Output {
```
mellowcroc: This function is error-prone because it can change the state of `nodes` and calling a different `min_depth_assumed_correct` value can lead to a different state change.

For example, using the first test of `test_root_diff_depths_with` as an example, let's say data1's length is 4 instead of 8. Calling `add_data(tree, data0, 0)` and then `add_data(tree, data1, data0.len())` will create a tree that is populated except for indices [3,7,14,15,28,29,30,31]. 
IMG_8787
In this situation, calling `tree.get_root(2)` will result in a different root than calling `tree.get_root(3)` because `tree.get_root(2)` will recalculate the nodes from height 2 and `tree.get_root(3)` will recalculate from height 3 and the code just uses the zero-array as the node value.

I think there are a couple ways to solve this:
1. we can check whether both child nodes are 0, and if they are, skip recalculating them (unless they are leaf nodes)
1a. optionally make `get_root` just a getter function and move the state change logic into a separate function (e.g. `recalcuate_nodes`)
2. calculate all affected nodes when calling `add_data` and change `get_root` to be just a getter function (this could potentially slow down `add_data` calls and have concurrency issues)

commitment_scheme/src/merkle/merkle_commitment_scheme.rs
mellowcroc: actually, how about using `?` to propagate the error upwards?

merkle/src/hash.rs
varunthakore: I think it could be possible for `Blake2s256Hasher` and `Keccak256Hasher`. But I am not sure if it can be done for `Poseidon3Hasher` because its implementation of the `Hasher` trait slightly differs from the other two.

merkle/src/lib.rs
varunthakore: The logic of Verifier and Prover is now separate, which is using the channel implementation.

commitment_scheme/src/merkle/hash.rs
varunthakore: Yes, the `Hasher` trait implementation is same for `Blake2s256Hasher` and `Keccak256Hasher`. I thought of making it generic for all hashes but the trait implementation of `Poseidon3Hasher` is different from the other two, so I am not sure if it can be made generic over all hashes.

commitment_scheme/src/merkle/mod.rs
varunthakore: I am using the same test with different test vectors and hashes. I have combined the tests which used the same test vectors to avoid duplicates.

commitment_scheme/src/merkle/mod.rs
varunthakore: Combined the tests which used the same test vectors to avoid duplicate instantiation.

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[derive(Debug)]
+pub struct Keccak256Hasher<F: PrimeField> {
+    _ph: PhantomData<F>,
+}
+
+impl<F: PrimeField> Hasher<F> for Keccak256Hasher<F> {
+    const DIGEST_NUM_BYTES: usize = 32;
+
+    type Output = Vec<u8>;
```
rot256: Can this be a `[u8; 32]`?

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+        let hash = Poseidon3::hash(&input_felts).into_bigint().to_bytes_be();
+
+        let mut array = [0u8; 32];
+        array[..hash.len()].copy_from_slice(&hash);
+        array.to_vec()
+    }
+
+    fn node(input: &[Self::Output]) -> Self::Output {
+        assert_eq!(input.len(), 2);
+        let input0 = Felt252::from_be_bytes_mod_order(&input[0]);
```
rot256: Careful: Ideally you should check that the input nodes are already reduced integer.

The problem is that the function as defined now is not collision-resistant, e.g.

```
node(a, b) = node(a + p, b)
```

In practice it's probably fine, I can't find an attack for the particular application.
But it does mean that e.g. Merkle tree openings are malleable.

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+
+// Implement Hasher trait for MaskedHash
+impl<
+        F: PrimeField,
+        H: Hasher<F, Output = Vec<u8>>,
+        const NUM_EFFECTIVE_BYTES: usize,
+        const IS_MSB: bool,
+    > Hasher<F> for MaskedHash<F, H, NUM_EFFECTIVE_BYTES, IS_MSB>
+{
+    const DIGEST_NUM_BYTES: usize = H::DIGEST_NUM_BYTES;
```
rot256, line 144: Should this be `NUM_EFFECTIVE_BYTES`?

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    type Output = Vec<u8>;
```
rot256: Could this be `[u8; NUM_EFFECTIVE_BYTES]`?

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+    type Output: Clone + Eq + Default + Debug + AsRef<[u8]>;
+
+    // compress a list of internal nodes into a single internal node
+    fn node(input: &[Self::Output]) -> Self::Output;
+
+    // compress a list of leaves into a single leaf
+    fn leaf(input: &[F]) -> Self::Output;
+
+    // compress a list of bytes into a single internal node
+    fn hash_bytes(data: &[u8]) -> Self::Output;
```
rot256, line 38: Somewhat inelegant to add this to the interface, but I don't have a better suggestion. From `hash_bytes` we can provide default implementations of `fn node` and `fn leaf`.

commitment_scheme/src/merkle/hash.rs
```diff
@@ -0,0 +1,273 @@
...
+
+#[derive(Debug)]
+pub struct Blake2s256Hasher<F: PrimeField> {
+    _ph: PhantomData<F>,
+}
+
+impl<F: PrimeField> Hasher<F> for Blake2s256Hasher<F> {
+    const DIGEST_NUM_BYTES: usize = 32;
+
+    type Output = Vec<u8>;
```
rot256: `type Output = [u8; 32]` ?

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+    fn add_segment_for_commitment(&mut self, segment_data: &[u8], segment_index: usize);
+
+    // Commit to the data
+    fn commit(&mut self);
+
+    // Start the decommitment phase
+    fn start_decommitment_phase(&mut self, queries: BTreeSet<usize>) -> Vec<usize>;
+
+    // Decommit to data stored in queried locations
+    fn decommit(&mut self, elements_data: &[u8]);
```
rot256: I don't understand the type of the method: are the queried locations serialized into some format?

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+    merkle_commitment_scheme::{MerkleCommitmentSchemeProver, MerkleCommitmentSchemeVerifier},
+};
+use packaging_commitment_scheme::{
+    PackagingCommitmentSchemeProver, PackagingCommitmentSchemeVerifier,
+};
+use randomness::Prng;
+use sha3::Digest;
+
+// Define the CommitmentSchemeProver trait
+pub trait CommitmentSchemeProver {
```
rot256: Just to clarify, this is not implemented?

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+    fn segment_length_in_elements(&self) -> usize;
+
+    // Return the size of an element, measured in bytes
+    fn element_length_in_bytes(&self) -> usize;
+
+    // Add a segment for commitment
+    fn add_segment_for_commitment(&mut self, segment_data: &[u8], segment_index: usize);
+
+    // Commit to the data
+    fn commit(&mut self);
```
rot256: Presumably this "finishes" the "add_segment_for_commitment" and starts the part where the prover can open positions, so it might make sense for it to take ownership of the struct and return a "Commitment" object with opening methods on it.

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+{
+    let n_layers = n_elements.ilog2() as usize;
+    let n_verifier_friendly_layers = std::cmp::min(n_layers, n_verifier_friendly_commitment_layers);
+
+    let mut cur_n_elements_in_layer = 1;
+    let is_top_hash = n_verifier_friendly_layers > 0;
+    let hash_name = commitment_hashes.get_hash_name(is_top_hash);
+
+    let mut next_inner_layer: Box<dyn CommitmentSchemeVerifier>;
+    match hash_name.as_str() {
```
rot256: Consider using an `enum` here, something like `SupportedHashes` and implement `FromStr`

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+            n_out_of_memory_merkle_layers,
+            n_elements_inner_layer / n_segments,
+            n_segments,
+            channel_clone,
+            n_verifier_friendly_commitment_layers,
+            commitment_hashes,
+        )
+    });
+
+    match hash_name.as_str() {
```
rot256: You can just return `match` expression, no need for assigning to `commitment_scheme`

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+    fn verify_integrity(&mut self, elements_to_verify: BTreeMap<usize, Vec<u8>>) -> Option<bool>;
+
+    // Return the total number of elements in the current layer
+    fn num_of_elements(&self) -> usize;
+}
+
+// Commitment hashes containing top and bottom hash strings
+#[derive(Debug, Clone)]
+pub struct CommitmentHashes {
+    top_hash: String,
```
rot256: Make it an enum.

commitment_scheme/src/merkle/merkle_commitment_scheme.rs
```diff
@@ -0,0 +1,373 @@
...
+
+    fn add_segment_for_commitment(&mut self, segment_data: &[u8], segment_index: usize) {
+        assert!(segment_data.len() == self.segment_length_in_elements() * self.size_of_element);
+        self.tree.add_data(
+            &bytes_as_hash::<F, H>(segment_data, self.size_of_element),
+            segment_index * self.segment_length_in_elements(),
+        );
+    }
+
+    fn commit(&mut self) {
```
rot256: Should this just take a 'Channel'? That way you don't need to store it, can simplify the type and might be able to avoid the `Rc<RefCell<..>>`?

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    ///
+    /// - `data_length`: length of data which the tree will commit to.
+    ///
+    /// # Returns
+    ///
+    /// - `Self`: MerkleTree with empty nodes
+    pub fn new(data_length: usize) -> Self {
+        // assert data_length is a power of 2
+        assert!(
+            data_length != 0 && (data_length & (data_length - 1)) == 0,
```
rot256: Use https://doc.rust-lang.org/std/primitive.usize.html#method.is_power_of_two instead for readability.

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    /// Constructs a new MerkleTree with empty nodes.
+    ///
+    /// # Arguments
+    ///
+    /// - `data_length`: length of data which the tree will commit to.
+    ///
+    /// # Returns
+    ///
+    /// - `Self`: MerkleTree with empty nodes
+    pub fn new(data_length: usize) -> Self {
```
rot256: Maybe change `data_length` to `num_leafs`

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+}
+
+impl<F: PrimeField, H: Hasher<F, Output = Vec<u8>>> MerkleTree<F, H> {
+    /// Feeds the tree with data to commit on.
+    ///
+    /// # Arguments
+    ///
+    /// - `data`: input data to commit on.
+    /// - `start_index`: index of the tree leaves where the data will be added.
+    pub fn add_data(&mut self, data: &[H::Output], start_index: usize) {
```
rot256, line 58: Is this ever used with internal nodes or only leafs?

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Returns an Option<bool> which is true if data_to_verify is present in the tree.
+    pub fn verify_decommitment<P: Prng, W: Digest>(
+        merkle_root: H::Output,
+        total_data_length: usize,
+        data_to_verify: &BTreeMap<usize, H::Output>,
+        channel: &mut FSVerifierChannel<F, P, W>,
+    ) -> Option<bool> {
```
rot256: Consider changing the return type to a `Result<(), Error>` which will allow us to use `?` from the caller.

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+/// - `size_of_element`: length of element in bytes.
+///
+/// # Returns
+///
+/// Returns vector of hash output.
+pub fn bytes_as_hash<F: PrimeField, H: Hasher<F, Output = Vec<u8>>>(
+    bytes_data: &[u8],
+    size_of_element: usize,
+) -> Vec<H::Output> {
+    let n_elements = bytes_data.len() / size_of_element;
```
rot256: Use this https://doc.rust-lang.org/std/slice/struct.ChunksExact.html instead.

commitment_scheme/src/merkle/mod.rs
varunthakore: In the implementation, the `get_root` function is only called with `min_depth_assumed_correct` equal to height of the tree, even though in the test cases they are testing for different values of  `min_depth_assumed_correct`.

Since it is only called with `min_depth_assumed_correct` equal to height, I was thinking to remove the `min_depth_assumed_correct` argument from the function and update the function to be as follows:

```
pub fn get_root(&mut self) -> H::Output {
    let height_correct = self.data_length;
    
    for j in (1..height_correct).rev() {
        self.nodes[j] = H::node(&[self.nodes[2 * j].clone(), self.nodes[2 * j + 1].clone()]);
    }
    
    self.nodes[1].clone()
}
```
The updated function computes the root starting from the leaf nodes, so the above case of computing root from different depths will not occur.

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+                ))
+            }
+            &_ => unreachable!(),
+        };
+    }
+    next_inner_layer
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn make_commitment_scheme_prover<F, P, W>(
```
rot256, line 373: I think I might agree with Clippy here, might make sense to introduce a "configuration" or "parameters" struct holding these.

commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,912 @@
...
+    size_of_element: usize,
```
rot256, line 374: Can this be derived from the field? e.g F::MODULUS_BIT_SIZE.div_ceil(8)

commitment_scheme/src/packaging_commitment_scheme.rs
```diff
@@ -0,0 +1,1843 @@
...
+        );
+
+        for i in 0..self.missing_element_queries.len() {
+            let start = i * self.size_of_element;
+            let end = start + self.size_of_element;
+            let bytes_to_send = &elements_data[start..end];
+
+            if self.is_merkle_layer {
+                let digest = bytes_to_send.to_vec();
+                let mut channel = self.channel.borrow_mut();
```
rot256: In general, I think I'm a fan of explicitly passing a &mut Channel to these types of functions: it's easier to read what is written to the transcript in which order and you can see which methods affect the channel.

In the interest of getting this mega commit in, I think it's fine for now.

commitment_scheme/src/packaging_commitment_scheme.rs
```diff
@@ -0,0 +1,1843 @@
...
+            size_of_element,
+            n_elements,
+            channel,
+            packer,
+            inner_commitment_scheme,
+            is_merkle_layer,
+        }
+    }
+
+    pub fn new_test(
```
rot256: `#[cfg(test)]`?

commitment_scheme/src/packer_hasher.rs
```diff
@@ -0,0 +1,297 @@
...
+/// and use these as the basic element for the tree.
+#[derive(Debug)]
+pub struct PackerHasher<F: PrimeField, H: Hasher<F>> {
+    pub size_of_element: usize,
+    pub n_elements_in_package: usize,
+    pub n_packages: usize,
+    _ph: PhantomData<(F, H)>,
+}
+
+impl<F: PrimeField, H: Hasher<F, Output = Vec<u8>>> PackerHasher<F, H> {
```
rot256: Can the `Output` be `[u8; N]` i.e.

```
impl<F: PrimeField, H: Hasher<F, Output = Vec<u8>>> PackerHasher<F, H> {
```

Or alternatively:

```
impl<O: AsRef<[u8]>, F: PrimeField, H: Hasher<F, Output = O>> PackerHasher<F, H> {
```

commitment_scheme/src/packer_hasher.rs
```diff
@@ -0,0 +1,297 @@
...
+    // If n_elements == 0 and data is empty, return an empty vector.
+    if n_elements == 0 && data.is_empty() {
+        return Vec::new();
+    }
+
+    let element_size = data.len() / n_elements;
+    let mut res = Vec::with_capacity(n_elements * H::DIGEST_NUM_BYTES);
+    let mut pos = 0;
+
+    for _ in 0..n_elements {
```
rot256: Consider using ChunksExact. Do we need to pass `n_elements` or can we calculate it?

commitment_scheme/src/packer_hasher.rs
```diff
@@ -0,0 +1,297 @@
...
+pub fn hash_elements<F: PrimeField, H: Hasher<F, Output = Vec<u8>>>(
+    data: &[u8],
+    n_elements: usize,
+) -> Vec<u8> {
+    // If n_elements == 0 and data is empty, return an empty vector.
+    if n_elements == 0 && data.is_empty() {
+        return Vec::new();
+    }
+
+    let element_size = data.len() / n_elements;
```
rot256, line 240: This has no relation to the Field really: consider computing `n_elements = data / size_of(F)`

commitment_scheme/src/packer_hasher.rs
```diff
@@ -0,0 +1,297 @@
...
+    data: &[u8],
+) -> Vec<u8> {
+    // If data is empty, return an empty vector.
+    if data.is_empty() {
+        return Vec::new();
+    }
+
+    // Each element in the next layer is a hash of 2 elements in the current layer.
+    let elements_to_hash_size = 2 * H::DIGEST_NUM_BYTES;
+    let n_elements_next_layer = data.len() / elements_to_hash_size;
```
rot256, line 273: Makes sure that this divides: i.e. data.len() is a multiple of `2 * H::DIGEST_NUM_BYTES;`

commitment_scheme/src/merkle/mod.rs
```diff
@@ -0,0 +1,793 @@
...
+    let mut bytes_as_hash = Vec::with_capacity(n_elements);
+
+    for hash_idx in 0..n_elements {
```
rot256: Use ChunksExact again :)

commitment_scheme/src/table_prover.rs
```diff
@@ -0,0 +1,180 @@
...
+        verify_all_columns_same_length(columns),
+        "The sizes of the columns must be the same."
+    );
+
+    let n_columns = columns.len();
+    let n_rows = get_num_rows(columns);
+    let element_size_in_bytes = F::MODULUS_BIT_SIZE.div_ceil(8) as usize;
+    let n_bytes_row = n_columns * element_size_in_bytes;
+
+    let mut serialization = vec![0u8; n_rows * n_bytes_row];
```
rot256: let mut serialization = Vec::with_capacity(n_rows * n_bytes_row)

commitment_scheme/src/table_prover.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    for row in 0..n_rows {
+        for (col, col_data) in columns.iter().enumerate().take(n_columns) {
+            let element_idx = row * n_columns + col;
+            let element_byte_idx = element_idx * element_size_in_bytes;
+            let element_bytes = &col_data[row].into_bigint().to_bytes_be();
+
+            serialization[element_byte_idx..element_byte_idx + element_size_in_bytes]
```
rot256: serialization.extend_from_slice(&col_data[row].into_bigint().to_bytes_be());

Should give you the same result? 
Without having to compute the offsets explicitly.

commitment_scheme/src/table_utils.rs
```diff
@@ -0,0 +1,78 @@
...
+    integrity_queries: &BTreeSet<RowCol>,
+) -> BTreeSet<usize> {
+    let mut all_query_rows = BTreeSet::new();
+    for query in data_queries.iter() {
+        all_query_rows.insert(query.get_row());
+    }
+    for query in integrity_queries.iter() {
+        all_query_rows.insert(query.get_row());
+    }
+    all_query_rows
```
rot256: You can collect into a `BTreeSet` (like for a Vec). So another option is:

```
iter::empty()
   .chain(data_queries)
   .chain(integrity_queries)
   .map(|q| q.get_row())
   .collect()
```

commitment_scheme/src/table_verifier.rs
```diff
@@ -0,0 +1,495 @@
...
+                if let Some((&row_col_next, field_element)) = all_rows_it.next() {
+                    assert_eq!(
+                        row_col_next.get_row(),
+                        cur_row,
+                        "Data skips to next row before finishing the current."
+                    );
+                    // Copy the field element bytes into the appropriate position in row_data.
+                    let field_bytes = field_element.into_bigint().to_bytes_be();
+                    row_data[pos..pos + element_size].copy_from_slice(&field_bytes);
+                    pos += element_size;
```
rot256: Avoid `pos`:


let mut row_data = Vec::with_capacity(self.n_columns * element_size);

...

row_data.extend_from_slice(&field_bytes);

commitment_scheme/src/table_verifier.rs
```diff
@@ -0,0 +1,495 @@
...
+            response.insert(query_loc, field_element[0]);
+        }
+
+        Ok(response)
+    }
+
+    /// Given indexed field elements, verify that these field elements are indeed the ones committed to
+    /// by the prover, against the commitment obtained by read_commitment().
+    #[allow(dead_code)]
+    fn verify_decommitment(&mut self, all_rows_data: &BTreeMap<RowCol, F>) -> Option<bool> {
```
rot256: Change the return type to `Result<(), Error>`

commitment_scheme/src/table_verifier.rs
```diff
@@ -0,0 +1,495 @@
...
+                        row_col_next.get_row(),
+                        cur_row,
+                        "Data skips to next row before finishing the current."
+                    );
+                    // Copy the field element bytes into the appropriate position in row_data.
+                    let field_bytes = field_element.into_bigint().to_bytes_be();
+                    row_data[pos..pos + element_size].copy_from_slice(&field_bytes);
+                    pos += element_size;
+                } else {
+                    panic!("Not enough columns in the map.");
```
rot256: Is this unreachable? Otherwise we should return an error.

poseidon/src/lib.rs
```diff
@@ -137,6 +140,80 @@ impl FieldHasher<Felt252> for Poseidon3<Felt252> {
...
+            }
+        }
+        bits
+    };
+    let big_int = <BigInt<4> as BigInteger>::from_bits_be(&bits);
+    assert!(big_int < Felt252::MODULUS);
+    Felt252::from_bigint(big_int).expect("conversion fail")
+}
+
+pub fn hex_to_vec(hex_str: &str) -> Vec<u8> {
```
rot256: Consider renaming, it is specifically Hex to 32 Bytes e.g. `hex_to_b32`

rot256 (changes requested):
Looks good, just a few comments. Also, I don't know if there is some way to avoid the boxed closures, seems slightly complicated / hard to follow.

But overall: it's get this thing merged!
 
Finally: in the future, if the feature permits it, try and break the PRs into smaller ones: reviewing +5000 lines is a slightly daunting task.

commitment_scheme/src/merkle/hash.rs
varunthakore: I have updated the function as follows:

https://github.com/varunthakore/pumice/blob/83072ea4b5807180d3493889afc912afade771ed/commitment_scheme/src/merkle/hash.rs#L100-L114

commitment_scheme/src/merkle/hash.rs
varunthakore: The hash digest contains `DIGEST_NUM_BYTES` where `NUM_EFFECTIVE_BYTES` are non-zero and the remaining bytes are masked with `0u8`.

For iterative hashing to compute the nodes (suppose `DIGEST_NUM_BYTES` = 32), two 32-byte masked hash digest are first hashed using some hasher like `Keccak256Hasher` and then masked to output the 32-byte masked hash digest. As shown in the following test:

https://github.com/varunthakore/pumice/blob/83072ea4b5807180d3493889afc912afade771ed/commitment_scheme/src/merkle/hash.rs#L217-L224

commitment_scheme/src/merkle/hash.rs
varunthakore: I have updated this to `[u8; DIGEST_NUM_BYTES]`.

commitment_scheme/src/merkle/mod.rs
varunthakore: It is used to add new leaves data but it does compute the internal nodes of the subtree corresponding to the added leaves data.

commitment_scheme/src/packaging_commitment_scheme.rs
varunthakore: Updated the argument of the function to take `channel: &mut FSVerifierChannel<F, P, W>` as input.

commitment_scheme/src/packaging_commitment_scheme.rs
varunthakore: I was using this function for testing the implementation but forgot to remove it, have updated it.

commitment_scheme/src/lib.rs
varunthakore: Just removed the attribute, seems to be fine with Clippy.

commitment_scheme/src/packer_hasher.rs
varunthakore: `n_elements` here is number of packages `n_packages`, this function divides the data into `n_packages` and then hashes each package. So we need to pass the number of packages as argument to the function.

commitment_scheme/src/packer_hasher.rs
varunthakore: Yes, this has no relation to the Field. The function partitions `data` bytes into packages of size `2 * H::DIGEST_NUM_BYTES` and then hashes each package.

commitment_scheme/src/packer_hasher.rs
varunthakore: The following assert before calling this functions ensures that it divides.

https://github.com/varunthakore/pumice/blob/7fbe77735322bd9235b449969fb03a1399b797dd/commitment_scheme/src/packer_hasher.rs#L66-L69

commitment_scheme/src/lib.rs
varunthakore: It is used to prepare the prover for the decommit function. For example, in case of `PackagingCommitmentSchemeProver` given the queries, it computes the indices of the other elements in a package corresponding to the input queries and then returns these element indices. 

The data corresponding to these indices needs to be opened along with the query data to verify the commitment.

commitment_scheme/src/lib.rs
varunthakore: It has been implemented for `MerkleCommitmentSchemeProver` and `PackagingCommitmentSchemeProver` structs.

commitment_scheme/src/lib.rs
varunthakore: In the current implementation, all the commit and decommit functions have been implemented for the same struct like `MerkleCommitmentSchemeProver`, which is similar to the Stone implementation. 

Returning a "Commitment" object, requires defining a new trait containing the decommit functions and then implementing the trait for the "Commitment" struct. Which can be done but the overall functionality remains the same i.e. rather having commit and decommit on a single prover struct, we will have two separate structs for prover and commitment. 

Do you suggest to make these changes ? or should I go ahead with the current implementation ?

commitment_scheme/src/lib.rs
varunthakore: `size_of_element` is the size of element in bytes corresponding to the data the commitment scheme will commit to, where data will be of length `size_of_element * n_elements`. It need not be size of the field, for example check the following test:

https://github.com/varunthakore/pumice/blob/3ffb0fd79e7fa48119bbb01981ac15cd2d204ce5/commitment_scheme/src/packaging_commitment_scheme.rs#L431-L466

commitment_scheme/src/merkle/hash.rs
rot256: Hmm, sorry I did not consider this earlier, but this is actually also a problem: we crash if someone gives us an invalid child during a Merkle tree membership.

Don't know of the most elegant way to avoid this.
Maybe reverting to the previous code would be better.
Sorry about the trouble.

Longer term, I think the best way would be to move the range check into the deserialization code: 
so we can fail with a error when fed an invalid child.
That requires this `Output` to be `Felt252`, maybe something like:

```
impl<F: PrimeField> Hasher<F> for Poseidon3Hasher<F> {
    const DIGEST_NUM_BYTES: usize = 32;

    type Output = Felt252;

    fn leaf(input: &[F]) -> Self::Output {
        let input_felts: Vec<Felt252> = input
            .iter()
            .map(|f| {
                let bytes = f.into_bigint().to_bytes_be();
                Felt252::from_be_bytes_mod_order(&bytes)
            })
            .collect();
		Poseidon3::hash(&input_felts)
    }

    fn node(input: &[Self::Output]) -> Self::Output {
		Poseidon3::pair(input[0], input[1])
    }

    fn hash_bytes(data: &[u8]) -> Self::Output {
		Poseidon3::hash_bytes_to_field(data)
    }
}
```

Which would probably necessitate that:

```
impl<F: PrimeField> Hasher<F> for Poseidon3Hasher<F> {
    type Output: ark_serialize::CanonicalDeserialize + ark_serialize::CanonicalDeserialize;
	
	...
}
```

So we can go from/to outputs and bytes.
But this requires a bigger refactor, so let's reconsider that later.

commitment_scheme/src/lib.rs
rot256: Yes basically: the type encodes the set of valid operations at different stages.

But this is fine.

commitment_scheme/src/packer_hasher.rs
rot256: Maybe I'm missing context, but if, e.g. 

- data.len() = 4 * H::DIGEST_NUM_BYTES + 1
- n_elements_in_package = 2
- size_of_element = H::DIGEST_NUM_BYTES

Then:

- n_elements_in_data = data.len() / size_of_element = 4
- n_packages = n_elements_in_data / n_elements_in_package = 2
- data.len() / n_packages = 2 * H::DIGEST_NUM_BYTES

So the check passes because the remainder data.len() % n_packages is discarded.

Due to rounding it should probably be:

```
assert_eq!(data.len(), n_packages * 2 * H::DIGEST_NUM_BYTES, "Data size is wrong.");
```

In general when checking conditions like this try to avoid division: division is not defined over the ring of integers so you run into issues like this.

commitment_scheme/src/packer_hasher.rs
rot256: Ah okay, carry on then :)

commitment_scheme/src/packaging_commitment_scheme.rs
rot256: Thanks!

commitment_scheme/src/lib.rs
rot256: Ah okay, thanks for the context!

rot256 (approved):

PR #6, opened 2024-08-03
Dev felt channel impl

jaehunkim:


jaehunkim:
I'm going to merge this without further review to facilitate integration with the Merkle tree hash work. We'll address any issues in follow-up PRs if necessary.

PR #7, opened 2024-08-07
fix(dependencies): pin generic-array version to 0.14.7

jaehunkim:
- Resolve version conflicts in dependency chain
- Ensure compatibility with sha3, digest, and crypto-common


jaehunkim:
Given that the changes are minimal, and this has been previously agreed upon with @varunthakore, I merge this directly.

PR #9, opened 2024-08-19
refactor: Replace GenericArray with AsRef<[u8]> and Vec<u8>

jaehunkim:
- The readability has slightly decreased as the commitment type changed to Vec<u8>. A better implementation might involve some restrictions on the Commitment return type.
- Modified the verifier channel to allow for variable-sized commit hash retrieval (20 and 32). There might be a better implementation for this.


varunthakore:
looks good to me !

randomness/src/lib.rs
```diff
@@ -18,5 +15,6 @@ pub trait Prng {
     fn prng_state(&self) -> Vec<u8>;
     fn hash_name() -> &'static str;
 
+    fn digest_size() -> usize;
```
jaehunkim, line 18: Maybe there's better way to treat default digest size (32).

channel/src/fs_verifier_channel.rs
```diff
@@ -148,24 +146,26 @@ impl<F: PrimeField, P: Prng, W: Digest> VerifierChannel for FSVerifierChannel<F,
...
         self.states.increment_commitment_count();
         self.states.increment_hash_count();
-        Ok(commitment)
+        Ok(bytes)
     }
 
-    fn recv_decommit_node(&mut self) -> Result<Self::Commitment, anyhow::Error> {
-        let size = std::mem::size_of::<Self::Commitment>();
-        let bytes = self.recv_bytes(size)?;
+    fn recv_decommit_node_default(&mut self) -> Result<Vec<u8>, anyhow::Error> {
```
jaehunkim, line 161: The output type changes from `Self::Commitment` to an unmarked `Vec<u8>`. 
To improve readability, consider wrapping the vector in an enum or struct.

PR #11, opened 2024-09-03
bugfix: fix bug caused by not handling mont form of stone

jaehunkim:
According to the stone code, there was a difference in implementation when using Keccak256 as PRNG versus using Poseidon3 as PRNG, specifically in whether to view the random bytes array as a value with Montgomery form `(aR)` applied or not. This code modifications are to reflect this.

@varunthakore  It's likely that we'll need to modify the test code related to the Merkle tree.


channel/src/test.rs
```diff
@@ -122,18 +115,9 @@ impl TestFixtures for Keccak256TestTypes {
...
-                Felt252::from_be_bytes_mod_order(&hex!(
-                    "060b22b317393ca6509829cf22b8379cd9607c232836ff64ac34f89d0cc6b679"
-                )),
-                // Mont form of 0x2f06b17e08bc409b945b951de8102653dc48a143b87d09b6c95587679816d02_Z
-                Felt252::from_be_bytes_mod_order(&hex!(
-                    "06225ab6b4b37edbb3bb2d694ff09fda59ae3daeafb23c93db11ddfb7511a59b"
-                )),
+                hex("0x7f097aaa40a3109067011986ae40f1ce97a01f4f1a72d80a52821f317504992"),
+                hex("0x18bcafdd60fc70e5e8a9a18687135d0bf1a355d9882969a6b3619e56bf2d49d"),
+                hex("0x2f06b17e08bc409b945b951de8102653dc48a143b87d09b6c95587679816d02"),
```
jaehunkim, line 120: Now the values are well-aligned with `noninteractive_channel_test.cc`

PR #12, opened 2024-09-04
feat: Add helper function for LDE calculation

jaehunkim:
This is part of the effort to break down various tasks implemented for FRI implementation into smaller PR.
The LDE test referenced [here](https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/algebra/lde/lde_test.cc#L103).


Cargo.toml
```diff
@@ -1,11 +1,11 @@
...
-members = ["felt", "merkle", "poseidon", "randomness", "channel"]
+members = ["felt", "merkle", "poseidon", "randomness", "channel", "fri"]
 resolver = "2"
 
 [workspace.dependencies]
 anyhow = "1.0.86"
-ark-ff = "0.4.2"
-ark-poly = "0.4.2"
+ark-ff = { git = "https://github.com/jaehunkim/algebra", branch = "v0.4.2-fix" }
+ark-poly = { git = "https://github.com/jaehunkim/algebra", branch = "v0.4.2-fix" }
```
jaehunkim: diff : https://github.com/arkworks-rs/algebra/compare/master...jaehunkim:algebra:master
When q is 2, the function `get_root_of_unity(n)` always returns None for any n greater than 1.
I have temporarily patched this issue for now.

```rust
let q_adicity = crate::utils::k_adicity(q, n);
let q_part = q.checked_pow(q_adicity)?;
let two_adicity = crate::utils::k_adicity(2, n);
let two_part = 2u64.checked_pow(two_adicity)?;

/// two_part == q_part == n, so two_part * q_part will be n squared
if n != two_part * q_part
    || (two_adicity > Self::TWO_ADICITY)
    || (q_adicity > small_subgroup_base_adicity)
{
    return None;
}
```

fri/src/stone_domain.rs
```diff
@@ -0,0 +1,109 @@
...
+            new_index += s;
+        }
+        i >>= 1;
+        s >>= 1;
+    }
+    domain.elements().nth(new_index).unwrap()
+}
+
+// change order of elements in domain
+pub fn change_order_of_elements_in_domain<F: FftField>(elements: &[F]) -> Vec<F> {
```
jaehunkim, line 17: For a multiplicative group of order 8:
Typically, the elements are ordered as g^0, g^1, g^2, g^3, ...
However, in Stone's method, they use two types of element ordering:
One is the normal ordering, and the other is where
elements are ordered as g^0, g^4, g^2, g^6, ...

Cargo.toml
rot256: This turned out to be caused by a mistake on our end:

See https://github.com/zksecurity/pumice/pull/15

Sorry about that :)

fri/src/lde.rs
```diff
@@ -0,0 +1,148 @@
...
+            ldes: vec![],
+            base,
+            reversed_order,
+        }
+    }
+
+    // Adds an evaluation on coset that was used to build the LDE.
+    // Future eval invocations will add the lde of that evaluation to the results.
+    pub fn add_eval(&mut self, evaluation: &[F]) {
+        assert!(
```
rot256: `assert_eq!` has nicer output.

fri/src/lde.rs
```diff
@@ -0,0 +1,148 @@
...
+    // The results are ordered according to the order that the LDEs were added.
+    pub fn eval(&self, offset: F) -> Vec<Vec<F>> {
+        let eval_domain = self.base.get_coset(offset).unwrap();
+        let mut evals: Vec<Vec<F>> = vec![];
+        for lde in self.ldes.iter() {
+            let evals_lde = eval_domain.fft(lde);
+            evals.push(evals_lde);
+        }
+        evals
+    }
```
rot256: Consider using the https://docs.rs/ark-poly/0.4.2/ark_poly/polynomial/univariate/struct.DensePolynomial.html abstraction.

Also this might be a good candidate for parallelizing, e.g. using https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.map

fri/src/stone_domain.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mut new_index = 0;
+        let mut index = i;
+        for base in mapping_vec.iter() {
+            if index & 1 != 0 {
+                new_index += base;
+            }
+            index >>= 1;
+        }
+        new_elements[new_index] = *element;
+    }
```
rot256: This is just https://doc.rust-lang.org/std/primitive.u32.html#method.reverse_bits ? I guess you would also need a shift, e.g. `index.reverse_bits() >> trailing_zeroes` ?

rot256: Also:

```rust
let mut new_elements = Vec::with_capacity(size)
for i in 0..size {
   new_elements.push(elements[translate_index(i)])
}
```

Might be slightly faster.

fri/src/stone_domain.rs
```diff
@@ -0,0 +1,109 @@
...
+    let mut i = index;
+    let mut s = domain.size() / 2;
+    let mut new_index = 0;
+    while i > 0 {
+        if i & 1 != 0 {
+            new_index += s;
+        }
+        i >>= 1;
+        s >>= 1;
+    }
```
rot256: This is the same logic flipping the index?

Maybe factor this into a function, taking the bit_length and the index.

rot256 (approved):
Looks good overall. 
Turns out we don't need to fix anything in `ark-ff`.

PR #13, opened 2024-09-05
feat: implement commitment phase of fri verifier

jaehunkim:


fri/src/stone_domain.rs
```diff
@@ -0,0 +1,89 @@
...
+) -> Vec<Radix2EvaluationDomain<F>> {
+    let mut current_offset = offset;
+    let mut domains = Vec::with_capacity(domain_size_log + 1);
+    for i in (0..=domain_size_log).rev() {
+        let domain = Radix2EvaluationDomain::<F>::new(1 << i)
+            .unwrap()
+            .get_coset(current_offset)
+            .unwrap();
+        domains.push(domain);
+        current_offset = current_offset * current_offset;
```
jaehunkim: In Stone, when creating domains for each layer, the offset is squared

https://github.com/starkware-libs/stone-prover/blob/ba85f37de3f8473525b7dc03de36cdf8f1567775/src/starkware/fft_utils/fft_bases.inl#L57

fri/src/folder.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[allow(dead_code)]
+impl MultiplicativeFriFolder {
+    pub fn next_layer_element_from_two_previous_layer_elements<F: FftField>(
+        f_x: &F,
+        f_minus_x: &F,
+        eval_point: &F,
+        x: &F,
+    ) -> F {
+        let x_inv = x.inverse().unwrap();
```
rot256: Is this applied to many elements? You might want to use this:

https://docs.rs/ark-ff/latest/ark_ff/fields/fn.batch_inversion.html

fri/src/folder.rs
```diff
@@ -0,0 +1,26 @@
...
+        Self::fold(f_x, f_minus_x, eval_point, &x_inv)
+    }
+
+    fn fold<F: FftField>(f_x: &F, f_minus_x: &F, eval_point: &F, x_inv: &F) -> F {
```
rot256: I would just take the field elements: you are deref'ing anyway.

fri/src/folder.rs
```diff
@@ -0,0 +1,26 @@
...
+        *f_x + *f_minus_x + *eval_point * (*f_x - *f_minus_x) * *x_inv
```
rot256: Maybe add comment about what this does: it is interpolating a line through (x, f(x)) and (-x, f(-x)) then evaluating it at "eval_point"

rot256: You can see this by setting eval_point = -x and x respectively :)

fri/src/folder.rs
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+
+#[allow(dead_code)]
+pub fn fri_folder_from_field() -> Arc<MultiplicativeFriFolder> {
```
rot256: Why do we need this?

Creating a struct with no field costs nothing, alternatively create a global constant instance.

fri/src/parameters.rs
```diff
@@ -0,0 +1,102 @@
...
+           non-negligible.
+        2. It requires N to be of the form 2^n.
+
+      In our implementation, we reduce the degree from N to R (last_layer_degree_bound) for a
+      relatively small R using log2(N/R) fri steps. To do it we send the R coefficients of the last
+      FRI layer instead of continuing with additional FRI layers.
+
+      To reduce proof-length, it is always better to pick last_layer_degree_bound > 1.
+    */
+    pub last_layer_degree_bound: u64,
```
rot256: Nitpick: I would make this a usize to.

fri/src/parameters.rs
```diff
@@ -0,0 +1,102 @@
...
+      of 2. Also note that in the fri_step_0=1 case we send 2 additional field elements per query (one
+      for each of the two layers that we skipped). So, while we send more field elements in the
+      fri_step_0=3 case (8 rather than 4), we refrain from sending the authentication paths for the
+      two skipped layers.
+
+      For a simple FRI usage, take fri_step_list = {1, 1, ..., 1}.
+    */
+    pub fri_step_list: Vec<usize>,
+
+    /*
```
rot256: Use `///` instead: that way it shows up in documentation as well :)

fri/src/parameters.rs
```diff
@@ -0,0 +1,102 @@
...
+            ph: PhantomData,
+            fri_step_list,
+            last_layer_degree_bound,
+            n_queries,
+            fft_domains,
+            proof_of_work_bits,
+        }
+    }
+
+    pub fn from_json(json: &JsonValue, fft_domains: Vec<E>) -> Self {
```
rot256: Manually doing this is *almost* always a bad idea: for instance, you are ignoring additional fields / garbage in the JSON object. Consider creating a separate struct and `#[derive(Deserialize)]` on it.

The function should probably return a result, rather than panic.

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+use sha3::Digest;
+
+use crate::{lde::MultiplicativeLDE, parameters::FriParameters};
+use channel::{fs_verifier_channel::FSVerifierChannel, Channel, VerifierChannel};
+use commitment_scheme::{
+    make_commitment_scheme_verifier, table_verifier::TableVerifier, CommitmentHashes,
+};
+
+#[allow(dead_code)]
+pub type FirstLayerQueriesCallback<F> = fn(&[u64]) -> Vec<F>;
```
rot256: What does this do? 

Might be cleaner to just add another trait parameter to the `FriVerifier`

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+            table_verifiers: vec![],
+            query_indices: vec![],
+            query_results: vec![],
+            expected_last_layer: None,
+            eval_points_test,
+            last_layer_coefficients_test,
+        }
+    }
+
+    pub fn verify_fri(&mut self) -> Result<(), Box<dyn Error>> {
```
rot256: Convert the error instead: handling a `dyn Error` seems like a pain.

Using anyhow is fine :)

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+                }
+            } else {
+                self.eval_points[i - 1] = self.channel.draw_felem();
+            }
+
+            if i < self.n_layers - 1 {
+                let coset_size = 1 << self.params.fri_step_list[i + 1];
+                let n_rows = self.params.fft_domains[basis_index].size() / coset_size;
+                let n_columns = coset_size;
+                let size_of_row = ((F::MODULUS_BIT_SIZE.div_ceil(8) * 8) as usize) * n_columns;
```
rot256: Factor this into a function:

```rust
fn byte_size<F: PrimeField>() -> usize {
   ...
}
```

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    pub fn commitment_phase(&mut self) -> Result<(), Box<dyn Error>> {
+        let mut basis_index = 0;
+        for i in 0..self.n_layers {
+            let cur_fri_step = self.params.fri_step_list[i];
+            basis_index += cur_fri_step;
+
+            if i == 0 {
+                if self.params.fri_step_list[0] != 0 {
+                    self.first_eval_point = Some(self.channel.draw_felem());
```
rot256: What's special about the first one?

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+                let commitment_scheme = make_commitment_scheme_verifier(
+                    size_of_row,
+                    n_rows,
+                    0,
+                    self.commitment_hashes.clone(),
+                    n_columns,
+                );
+
+                let mut table_verifier = TableVerifier::new(n_columns, commitment_scheme);
+                let _ = table_verifier.read_commitment(&mut self.channel);
```
rot256, line 103: Maybe `make_commitment_scheme_verifier` should take the commitment: is there any scenario where it makes sense to create a verifier without knowing which commitment it verifies against?

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+        );
+
+        let last_layer_basis_index = fri_step_sum;
+        let lde_domain = self.params.fft_domains[last_layer_basis_index];
+        let mut lde = MultiplicativeLDE::new(lde_domain, true);
+
+        lde.add_coeff(&last_layer_coefficients_vector);
+        let evals = lde.eval(lde_domain.element(0));
+        self.expected_last_layer = Some(evals[0].clone());
+
```
rot256: Wait, the query point of the last layer is always g^0 ?

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+        let last_layer_size = self.params.fft_domains[fri_step_sum].size();
+
+        let mut last_layer_coefficients_vector = if let Some(last_layer_coefficients_test) =
+            self.last_layer_coefficients_test.as_ref()
+        {
+            last_layer_coefficients_test.clone()
+        } else {
+            self.channel
+                .recv_felts(self.params.last_layer_degree_bound as usize)
+                .unwrap()
```
rot256: Return an error instead: the prover cannot be trusted to provide a correctly encoded proof, we want to fail, not to crash.

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+        self.read_last_layer_coefficients()?;
+        Ok(())
+    }
+
+    pub fn read_last_layer_coefficients(&mut self) -> Result<(), Box<dyn Error>> {
+        let fri_step_sum: usize = self.params.fri_step_list.iter().sum();
+        let last_layer_size = self.params.fft_domains[fri_step_sum].size();
+
+        let mut last_layer_coefficients_vector = if let Some(last_layer_coefficients_test) =
+            self.last_layer_coefficients_test.as_ref()
```
rot256: When does this path occur?

It seems to me that the prover would need to know the final folded polynomial before starting the FRI folding phase?

fri/src/verifier.rs
```diff
@@ -0,0 +1,282 @@
...
+        } else {
+            self.channel
+                .recv_felts(self.params.last_layer_degree_bound as usize)
+                .unwrap()
+        };
+
+        // pad last_layer_coefficients_vector with zeros to the size of last_layer_size
+        while last_layer_coefficients_vector.len() < last_layer_size {
+            last_layer_coefficients_vector.push(F::zero());
+        }
```
rot256, line 121: Should the prover not send something of size exactly "dimension of the final RS code"?

rot256 (commented):
I have a couple questions, sorry if I'm missing some context.

fri/src/folder.rs
jaehunkim: We don't need this.

fri/src/verifier.rs
jaehunkim: Will replace similar codes also.

fri/src/folder.rs
jaehunkim: will fix this on next PR

fri/src/parameters.rs
jaehunkim: Why do we need separate struct?
Seems a bit dumb question but, 
`#[derive(Deserialize)]` isn't enough?

fri/src/verifier.rs
jaehunkim: Nothing, but will change this in next PR

fri/src/verifier.rs
jaehunkim: Nope. `lde.eval()` was a batch evaluation where `lde` struct can have multiple polynomials. So `evals[0]` just means a first evaluation. not evaluation from g^0.

renamed `lde.eval()` to `lde.batch_eval()`

fri/src/verifier.rs
jaehunkim: It's confusing because TableVerifier uses 2-step initialization. Agreed.

fri/src/verifier.rs
jaehunkim: This is not included in the proof itself, it is derived from parameters. This is kinda 'known setup'

fri/src/verifier.rs
jaehunkim: Maybe implementation below would be better?
(nit: is it an anti-pattern returning Resut<Self, Error> in ctor?)

```rust

    pub fn new(
        n_columns: usize,
        commitment_scheme: Box<dyn CommitmentSchemeVerifier<F, P, W>>,
        channel: &mut FSVerifierChannel<F, P, W>,
    ) -> Result<Self, anyhow::Error> {
        let mut verifier = Self {
            n_columns,
            commitment_scheme,
        };
        verifier.commitment_scheme.read_commitment(channel)?;
        Ok(verifier)
    }
```

from 

```rust
    /// Create new TableVerifier.
    pub fn new(
        n_columns: usize,
        commitment_scheme: Box<dyn CommitmentSchemeVerifier<F, P, W>>,
    ) -> Self {
        Self {
            n_columns,
            commitment_scheme,
        }
    }

    /// Reads the initial commitment into the scheme (e.g., Merkle root).
    pub fn read_commitment(
        &mut self,
        channel: &mut FSVerifierChannel<F, P, W>,
    ) -> Result<(), anyhow::Error> {
        self.commitment_scheme.read_commitment(channel)
    }
```

rot256:
Are we ready to merge this PR?

PR #14, opened 2024-09-07
feat: implement decommitment phase of fri verifier

jaehunkim:


fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    for &idx in query_indices {
+        for i in (idx * first_layer_coset_size as u64)..((idx + 1) * first_layer_coset_size as u64)
+        {
+            first_layer_queries.push(i);
+        }
+    }
+
+    first_layer_queries
+}
```
rot256, line 29: Alternatively:

```rust
pub fn second_layer_queries_to_first_layer_queries(
    query_indices: &[u64],
    first_fri_step: usize,
) -> Vec<u64> {
    let first_layer_coset_size = (1 << first_fri_step) as u64;
    query_indices
         .iter()
         .copied()
         .map(|idx| (idx * first_layer_coset_size)..((idx + 1) * first_layer_coset_size))
         .flatten()
         .collect()
}
```

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+    eval_point: Option<F>,
+    params: &FriParameters<F, E>,
+    layer_num: usize,
+    mut first_element_index: usize,
+) -> F {
+    let mut curr_eval_point = eval_point;
+    let mut cumulative_fri_step = 0;
+    for i in 0..layer_num {
+        cumulative_fri_step += params.fri_step_list[i];
+    }
```
rot256, line 46:
```rust
let cumulative_fri_step = params.fri_step_list.take(layer_num).sum()
```

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+        let fft_domain = params.fft_domains[basis_index];
+
+        let mut next_layer = Vec::with_capacity(cur_layer.len() / 2);
+        for j in (0..cur_layer.len()).step_by(2) {
+            let res = MultiplicativeFriFolder::next_layer_element_from_two_previous_layer_elements(
+                &cur_layer[j],
+                &cur_layer[j + 1],
+                &curr_eval_point.unwrap(),
+                &get_field_element_at_index(&fft_domain, first_element_index + j),
+            );
```
rot256: Also mentioned in another PR: use batch inversion here.

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+    assert_eq!(
+        elements.len(),
+        1 << layer_fri_step,
+        "Number of elements is not consistent with the fri_step parameter."
+    );
+
+    let mut cur_layer = elements.to_vec();
+    for basis_index in cumulative_fri_step..(cumulative_fri_step + layer_fri_step) {
+        assert!(
+            curr_eval_point.is_some(),
```
rot256, line 58: Unwrap this outside the loop: nothing makes sense if it is None.

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+            first_layer_queries.push(i);
+        }
+    }
+
+    first_layer_queries
+}
+
+pub fn apply_fri_layers<F: FftField + PrimeField, E: EvaluationDomain<F>>(
+    elements: &[F],
+    eval_point: Option<F>,
```
rot256, line 37: Does `eval_point = None` make sense?

fri/src/details.rs
rot256: Actually, not even necessary: you are repeatedly inverting `curr_eval_point` just do this once :)

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut cur_layer = elements.to_vec();
+    for basis_index in cumulative_fri_step..(cumulative_fri_step + layer_fri_step) {
+        assert!(
+            curr_eval_point.is_some(),
+            "evaluation point doesn't have a value"
+        );
+        let fft_domain = params.fft_domains[basis_index];
+
+        let mut next_layer = Vec::with_capacity(cur_layer.len() / 2);
+        for j in (0..cur_layer.len()).step_by(2) {
```
rot256, line 64: Possible candidate for rayon par_map

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+    where
+        C: Fn(&[u64]) -> Vec<F> + 'static,
+    {
         let n_layers = params.fri_step_list.len();
         Self {
             channel,
             params,
             commitment_hashes,
-            first_layer_callback,
+            first_layer_callback: Box::new(first_layer_callback),
```
rot256: What does this callback do?

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+                    .unwrap()
+            };
+
+            for j in 0..self.query_results.len() {
+                let query_index = self.query_indices[j] >> (basis_index - first_fri_step);
+                let query_loc = get_table_prover_row_col(query_index, cur_fri_step);
+                to_verify.insert(query_loc, self.query_results[j]);
+            }
+
+            let eval_point = self.eval_points[i];
```
rot256, line 209: These eval_points are the FRI folding challenges right?

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+                let coset_start = get_table_prover_row(
+                    self.query_indices[j] >> (basis_index - first_fri_step),
+                    cur_fri_step,
+                );
+
+                for k in 0..coset_size {
+                    coset_elements.push(*to_verify.get(&RowCol::new(coset_start, k)).unwrap());
+                }
+
+                self.query_results[j] = apply_fri_layers(
```
rot256, line 222: Does this not refold everything for every level? i.e. n^2 behavior in the number of levels (n)? n is small (log(N)) but still. i.e. in the first call you fold 1 level, in the second you fold 2 levels, ...

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+        for i in (idx * first_layer_coset_size as u64)..((idx + 1) * first_layer_coset_size as u64)
+        {
+            first_layer_queries.push(i);
+        }
+    }
+
+    first_layer_queries
+}
+
+pub fn apply_fri_layers<F: FftField + PrimeField, E: EvaluationDomain<F>>(
```
rot256, line 35: Might be cleaner to have a method applying a single round of FRI folding, then apply it many times. The current function might hide unintended O(n^2) behavior.

fri/src/folder.rs
```diff
@@ -15,7 +15,7 @@ impl MultiplicativeFriFolder {
         Self::fold(f_x, f_minus_x, eval_point, &x_inv)
```
rot256: Calculate x_inv once and reuse over-and-over, currently computing this one-off inversion should dominate the folding cost.

rot256: Consider batch inverting all the x's across the FRI layers as well.

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
         let n_layers = params.fri_step_list.len();
         Self {
             channel,
             params,
             commitment_hashes,
-            first_layer_callback,
+            first_layer_callback: Box::new(first_layer_callback),
             n_layers,
             first_eval_point: None,
             eval_points: vec![F::zero(); n_layers - 1],
```
rot256: Can this just be an empty Vec to avoid "sentinel" values: better to crash with out-of-bounds that accidentally use a zero challenge without knowing it

fri/src/details.rs
rot256: This is not specific to `second_layer_queries_to_first_layer_queries` it computes the query positions for a set of cosets. Consider adding a comment explaining this behavior.

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+                continue;
+            } else {
+                assert!(
+                    self.table_verifiers[i]
+                        .verify_decommitment(&mut self.channel, &to_verify)
+                        .unwrap(),
+                    "Layer {} failed decommitment",
+                    i
+                );
+            }
```
rot256: It's really hard for me to follow when something is allowed to be a None/Some in this code.

If you expect things to be Some then consider unwrapping them at the top of the function: so that there is a clear set of preconditions for invoking the function.

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+                test_data[i].clone()
+            } else {
+                self.table_verifiers[i]
+                    .query(
+                        &mut self.channel,
+                        &layer_data_queries,
+                        &layer_integrity_queries,
+                    )
+                    .unwrap()
+            };
```
rot256: What does these two paths correspond to?

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+    for _ in 0..n_queries {
+        let random_index = channel.draw_number(domain_size as u64);
+        query_indices.push(random_index);
+    }
+
+    query_indices.sort_unstable();
+    query_indices
+}
+
+pub fn next_layer_data_and_integrity_queries<F: FftField + PrimeField, E: EvaluationDomain<F>>(
```
rot256, line 116: Maybe a comment explaining what these sets are.

fri/src/details.rs
```diff
@@ -0,0 +1,142 @@
...
+            idx >> cumulative_fri_step,
+            layer_fri_step,
+        ));
+    }
+
+    for &idx in query_indices {
+        let coset_row = get_table_prover_row(idx >> cumulative_fri_step, layer_fri_step);
+        for coset_col in 0..(1 << layer_fri_step) {
+            let query = RowCol::new(coset_row, coset_col);
+            if !integrity_queries.contains(&query) {
```
rot256, line 138: Can `integrity_queries.contains(&query)` occur?

fri/src/verifier.rs
```diff
@@ -39,48 +49,173 @@ pub struct FriVerifier<
...
+                    &coset_elements,
+                    Some(eval_point),
+                    &self.params,
+                    i + 1,
+                    coset_start * (1 << cur_fri_step),
+                );
+            }
+
+            if self.query_indices_test.is_some() {
+                continue;
```
rot256: This "continue" does nothing: there is no logic following the clause.

rot256 (commented):
I find this code somewhat hard to follow:

There is a lot of functions mutating state on the same large object, lots of values which can be either None/Some which are set/unwrapped inside nested loops, lots of indexing. It's really hard to discern which state is being update by what and which state is relevant to a particular function.

Might even make sense to merge the logic across `verify_first_layer()`, `verify_inner_layers()`, `verify_last_layer()`: currently they are transferring state between these functions via the `FriVerifier` struct anyway and it does not make sense to invoke one without the other, or even invoke them out-of-order: they are really "one transformation" on the object.

Consider adding some comments to the folding/verifier logic.

It also seems to me that it folds the same layer multiple times.

fri/src/details.rs
jaehunkim: Deleted this.

fri/src/details.rs
jaehunkim: Nope!

PR #15, opened 2024-09-11
fix: incorrect constants in Felt252 definition

rot256:
Turns out, we should not specify:

```
#[small_subgroup_base = "2"]
#[small_subgroup_power = "192"]
```

I believe these are used to interpolate over a coset if your field has insufficient 2-adicity.
Which is definitely not a problem for Felt252. This was causing issues with the LDE impl.


jaehunkim (approved):
LGTM!

PR #16, opened 2024-09-11
Implement compute_next_fri_layer and add tests

varunthakore:
This PR implements `compute_next_fri_layer` function and add tests corresponding to the following Stone file.

https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/fri/fri_details_test.cc

### DIFF
--- a/fri/src/folder.rs
+++ b/fri/src/folder.rs
@@ -1,9 +1,9 @@
+use crate::stone_domain::get_field_element_at_index;
 use anyhow::Error;
+use ark_ff::fields::batch_inversion;
 use ark_ff::{FftField, PrimeField};
 use ark_poly::Radix2EvaluationDomain;
 use std::sync::Arc;
-
-use crate::stone_domain::get_field_element_at_index;
 
 pub struct MultiplicativeFriFolder;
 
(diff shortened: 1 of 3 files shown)


fri/src/folder.rs
```diff
@@ -1,10 +1,41 @@
...
+        input_layer: &[F],
+        eval_point: &F,
+    ) -> Result<Vec<F>, Error> {
+        assert_eq!(input_layer.len(), domain.size as usize);
+
+        let mut next_layer = Vec::with_capacity(input_layer.len() / 2);
+        for j in (0..input_layer.len()).step_by(2) {
+            let x_inv = get_field_element_at_index(&domain, j)
+                .inverse()
+                .ok_or_else(|| anyhow::anyhow!("Inverse not found"))?;
```
rot256: Same comment about using batch inversion, or, just batch precompute all the "x_inv".
In the case of the domain, computing the inverse is pretty easy: the domain is a (coset of) a cyclic subgroup, i.e. so the inverse is just the element at index j + n/2.

rot256 (approved):
LGTM!

PR #18, opened 2024-09-23
Refactor Table Prover and Fix endianness

varunthakore:
This PR make the following changes:
- minor changes to `TableProver` constructor
- change field element to Montgomery form before feeding into channel
- fix `seed_increment` byte representation from `to_le_bytes` to `to_be_bytes` in `mix_seed_with_bytes`

### DIFF
--- a/commitment_scheme/src/table_prover.rs
+++ b/commitment_scheme/src/table_prover.rs
@@ -1,5 +1,6 @@
+use crate::make_commitment_scheme_prover;
 use crate::table_utils::{all_query_rows, elements_to_be_transmitted, RowCol};
-use crate::CommitmentSchemeProver;
+use crate::{CommitmentHashes, CommitmentSchemeProver};
 use ark_ff::{BigInteger, PrimeField};
 use channel::fs_prover_channel::FSProverChannel;
 use channel::ProverChannel;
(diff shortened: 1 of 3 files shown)


commitment_scheme/src/table_prover.rs
```diff
@@ -163,7 +187,8 @@ fn serialize_field_columns<F: PrimeField>(segment: &[Vec<F>]) -> Vec<u8> {
 
     for row in 0..n_rows {
         for col_data in columns.iter().take(n_columns) {
-            serialization.extend_from_slice(&col_data[row].into_bigint().to_bytes_be());
+            let data_mont = col_data[row] * mont_r;
```
rot256, line 190: What is `mont_r`, can it be computed from the field instead of passed?

rot256 (approved):
No comments on my end.

Looks good!

commitment_scheme/src/table_prover.rs
varunthakore: It is the factor used to convert the field element from Standard to Montgomery form. It is used at several places, so just computing it once and passing as argument to other functions.

varunthakore:
Closing this PR since merged #23 directly into master.

PR #19, opened 2024-09-23
Implement FRI committed layers

varunthakore:
Implements FRI layers and committed layers corresponding to the following Stone library:

https://github.com/starkware-libs/stone-prover/tree/main/src/starkware/fri

This PR has dependency on #18.

### DIFF
--- a/fri/src/layers.rs
+++ b/fri/src/layers.rs
@@ -1,0 +1,378 @@
+use crate::folder::MultiplicativeFriFolder;
+use anyhow::Error;
+use ark_ff::PrimeField;
+use ark_poly::EvaluationDomain;
+use std::sync::Arc;
+
+#[allow(dead_code)]
+pub trait FriLayer<F: PrimeField, E: EvaluationDomain<F>> {
+    fn get_layer_size(&self) -> usize;
+    fn get_domain(&self) -> E;
+    fn get_layer(&self) -> Result<Vec<F>, Error>;
+    fn eval_at_points(&self, required_indices: &[usize]) -> Vec<F>;
+    fn clone_arc(&self) -> Arc<dyn FriLayer<F, E>>;
+}
+
+#[derive(Clone)]
+pub struct FriLayerReal<F: PrimeField, E: EvaluationDomain<F>> {
+    layer_size: usize,
+    domain: E,
+    evaluation: Vec<F>,
+}
+
+#[allow(dead_code)]
+impl<F: PrimeField, E: EvaluationDomain<F>> FriLayerReal<F, E> {
+    pub fn new(domain: E, evaluation: Vec<F>) -> Self {
+        Self {
+            layer_size: domain.size(),
+            domain,
+            evaluation,
+        }
+    }
+
+    pub fn new_from_prev_layer(prev_layer: &dyn FriLayer<F, E>) -> Self {
+        Self {
+            layer_size: prev_layer.get_layer_size(),
+            domain: prev_layer.get_domain(),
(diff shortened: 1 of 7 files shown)


fri/src/layers.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    fn eval_at_points(&self, required_indices: &[usize]) -> Vec<F> {
+        required_indices
+            .iter()
+            .map(|&i| self.evaluation[i])
+            .collect()
+    }
+
+    fn clone_arc(&self) -> Arc<dyn FriLayer<F, E>> {
```
rot256, line 62: Why is this method needed? Can't the user just wrap the thing in an Arc himself?

This also copies the underlying struct, I would imagine that the goal was to have a number of references to the *same* immutable instance?

fri/src/layers.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    fn eval_at_points(&self, _required_indices: &[usize]) -> Vec<F> {
+        unimplemented!("Should never be called")
+    }
+
+    fn clone_arc(&self) -> Arc<dyn FriLayer<F, E>> {
+        Arc::new(FriLayerProxy {
+            layer_size: self.layer_size,
+            domain: self.domain,
+            prev_layer: Arc::clone(&self.prev_layer),
```
rot256, line 122: `self.prev_layer.clone()` does the same.

fri/src/layers.rs
```diff
@@ -0,0 +1,378 @@
...
+    fn get_layer_size(&self) -> usize;
+    fn get_domain(&self) -> E;
+    fn get_layer(&self) -> Result<Vec<F>, Error>;
+    fn eval_at_points(&self, required_indices: &[usize]) -> Vec<F>;
+    fn clone_arc(&self) -> Arc<dyn FriLayer<F, E>>;
+}
+
+#[derive(Clone)]
+pub struct FriLayerReal<F: PrimeField, E: EvaluationDomain<F>> {
+    layer_size: usize,
```
rot256, line 18: `layer_size` is a redundant field: it looks to be the same as `domain.size()`?

fri/src/layers.rs
```diff
@@ -0,0 +1,378 @@
...
+impl<F: PrimeField, E: EvaluationDomain<F>> FriLayerReal<F, E> {
+    pub fn new(domain: E, evaluation: Vec<F>) -> Self {
+        Self {
+            layer_size: domain.size(),
+            domain,
+            evaluation,
+        }
+    }
+
+    pub fn new_from_prev_layer(prev_layer: &dyn FriLayer<F, E>) -> Self {
```
rot256, line 33: This just clones the layer?

rot256 (approved):
A couple questions, otherwise LGTM.

fri/src/layers.rs
varunthakore: Yes, removed this function.

fri/src/layers.rs
varunthakore: Removed the `clone_arc` function.

fri/src/layers.rs
varunthakore: Removed `layer_size`.

fri/src/layers.rs
varunthakore: It is used to create `FriLayerReal` from the previous `FriLayerProxy`, both the structs implement `FriLayer` trait.

varunthakore:
Closing this PR since #24 is merged.

PR #22, opened 2024-09-30
Implement fri verifier

jaehunkim:


commitment_scheme/src/table_verifier.rs
```diff
@@ -85,8 +100,13 @@ impl<F: PrimeField, P: Prng, W: Digest> TableVerifier<F, P, W> {
...
                         "Data skips to next row before finishing the current."
                     );
+
                     // Copy the field element bytes into the appropriate position in row_data.
-                    let field_bytes = field_element.into_bigint().to_bytes_be();
+                    let field_bytes = if self.convert_mont_decommitment {
+                        field_element.mul(&self.mont_r).into_bigint().to_bytes_be()
+                    } else {
+                        field_element.into_bigint().to_bytes_be()
+                    };
```
jaehunkim, line 109: @varunthakore While writing tests for the FRI verifier, it seems that this task is necessary to synchronize with stone-prover. Is there perhaps a better method?

varunthakore: These changes look good !

PR #25, opened 2024-10-09
feat: implement Pedersen hash

varunthakore:
This PR implements the Pedersen hash used in Stone.
https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/crypt_tools/pedersen.inl

It might seem like a huge PR but most of it is just constants.


pedersen/src/hash_context.rs
```diff
@@ -0,0 +1,5153 @@
...
+        n_hash_inputs,
+        points[0],
+        points[2..(2 + n_hash_inputs * element_bits_hash)].to_vec(),
+    )
+}
+
+// Copied from https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/algebra/elliptic_curve/elliptic_curve_constants.h
+pub fn get_points() -> Vec<Affine> {
+    vec![
+        Affine::new(
```
rot256: I did not really intent for felt::hex to be used outside of tests, is it possible to do this https://github.com/zksecurity/pumice/blob/master/poseidon/src/constants.rs#L21 instead? Then you can also just define it as a constant array.

pedersen/src/hash_context.rs
```diff
@@ -0,0 +1,5153 @@
...
+        element_bits_hash,
+        ec_subset_sum_height,
+        n_hash_inputs,
+        points[0],
+        points[2..(2 + n_hash_inputs * element_bits_hash)].to_vec(),
+    )
+}
+
+// Copied from https://github.com/starkware-libs/stone-prover/blob/main/src/starkware/algebra/elliptic_curve/elliptic_curve_constants.h
+pub fn get_points() -> Vec<Affine> {
```
rot256: Consider moving these into separate file, something like `constants.rs`

pedersen/src/hash_context.rs
```diff
@@ -0,0 +1,5153 @@
...
+    points: Vec<Affine>,
+}
+
+impl PedersenHashContext {
+    pub fn new(
+        n_element_bits: usize,
+        ec_subset_sum_height: usize,
+        n_inputs: usize,
+        shift_point: Affine,
+        points: Vec<Affine>,
```
rot256: Do we use any other set of points? Otherwise, might make sense to just hardcode them to simplify the interface.

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut bits = Vec::new();
+        for byte in bytes {
+            for i in (0..8).rev() {
+                bits.push(byte & (1 << i) != 0);
+            }
+        }
+        bits
+    };
+    let big_int = <BigInt<4> as BigInteger>::from_bits_be(&bits);
+    assert!(big_int < Felt252::MODULUS);
```
rot256: This check is important.

However an incomplete function should return a sum-type:
Return an Option/Result and handle the error case.

We don't want to potentially crash if someone feeds us an invalid hash.

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+fn bytes_to_field(bytes: &[u8]) -> Felt252 {
+    let bits = {
+        let mut bits = Vec::new();
+        for byte in bytes {
+            for i in (0..8).rev() {
+                bits.push(byte & (1 << i) != 0);
+            }
+        }
+        bits
+    };
```
rot256, line 68: Should probably also check that the bits fit inside a BigInt with 4 limbs: otherwise a hash has a non-canonical encoding.

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+            state: bytes_to_field(&digest),
+        }
+    }
+
+    pub fn get_digest(&self) -> Vec<u8> {
+        self.state.into_bigint().to_bytes_be()
+    }
+
+    pub fn hash(val0: PedersenHash, val1: PedersenHash) -> PedersenHash {
+        let cntx = get_standard_pedersen_hash_context();
```
rot256: Either see if you can create a constant for the context or, if not possible, consider lazy_static

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    pub fn hash_bytes_with_length(bytes: Vec<u8>) -> PedersenHash {
+        let mut state = Felt252::ZERO;
+        let cntx = get_standard_pedersen_hash_context();
+        let modulus = BigUint::from_bytes_be(&Felt252::MODULUS.to_bytes_be());
+
+        let mut bytes_to_hash = bytes.len();
+        let mut offset = 0;
+
+        while bytes_to_hash >= 32 {
```
rot256: Use https://doc.rust-lang.org/std/slice/struct.Chunks.html

Move the value "32" into a constant with a comment, "PEDERSEN_CHUNK_SIZE" maybe?

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn hash(val0: PedersenHash, val1: PedersenHash) -> PedersenHash {
+        let cntx = get_standard_pedersen_hash_context();
+        let res = cntx.hash_elements(val0.state, val1.state);
+        PedersenHash { state: res }
+    }
+
+    pub fn hash_bytes_with_length(bytes: Vec<u8>) -> PedersenHash {
+        let mut state = Felt252::ZERO;
+        let cntx = get_standard_pedersen_hash_context();
+        let modulus = BigUint::from_bytes_be(&Felt252::MODULUS.to_bytes_be());
```
rot256, line 33: https://docs.rs/ark-ff/latest/ark_ff/fields/models/fp/struct.Fp.html#method.into_bigint

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+        self.state.into_bigint().to_bytes_be()
+    }
+
+    pub fn hash(val0: PedersenHash, val1: PedersenHash) -> PedersenHash {
+        let cntx = get_standard_pedersen_hash_context();
+        let res = cntx.hash_elements(val0.state, val1.state);
+        PedersenHash { state: res }
+    }
+
+    pub fn hash_bytes_with_length(bytes: Vec<u8>) -> PedersenHash {
```
rot256: Trying to understand this code: it divides the next chunk by the *base field* of the curve? Should it not be the scalar field of the curve?

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        while bytes_to_hash >= 32 {
+            let word = BigUint::from_bytes_be(&bytes[offset..offset + 32]);
+            let q = word.clone() / modulus.clone();
+            let r = word % modulus.clone();
+            let value = bytes_to_field(&r.to_bytes_be());
+            assert!(q < BigUint::from(1000u64), "Unexpectedly large shift.");
+            let shift = bytes_to_field(&q.to_bytes_be());
+
+            state = cntx.hash_elements(state, value) + shift;
```
rot256, line 46: I honestly don't understand the types here:

- `state` is a field element of the basefield
- `value` is a field element of the basefield

Value and State are then going to be decomposed and used as scalars on the curve.
Finally the x-coordinate is taken as the hash (after adding the shift point).

But then this "shift" (the quotient) is added in the field? No curve operation?

I'm trying to reconcile it with https://docs.starkware.co/starkex/crypto/pedersen-hash-function.html

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+
+            offset += 32;
+            bytes_to_hash -= 32;
+        }
+
+        assert!(
```
rot256: nitpick: assert_eq

rot256: Move the assert to the top, to make it clear that it is a precondition of the function body: fail early if the precondition is not satisfied.

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+        let input = vec![];
+        let res = PedersenHash::hash_bytes_with_length(input);
+        let exp_res =
+            felt::hex("0x49ee3eba8c1600700ee1b87eb599f16716b0b1022947733551fde4050ca6804");
+        assert_eq!(exp_res, res.state);
+    }
+
+    #[test]
+    fn test_pedersen() {
+        let input = generate_test_vector(32);
```
rot256: Use (multiples of) the constant: also conveys that only "full blocks" are supported.

pedersen/src/pedersen.rs
```diff
@@ -0,0 +1,118 @@
...
+use ark_ff::{BigInt, BigInteger, Field, PrimeField};
+use felt::Felt252;
+use num_bigint::BigUint;
+
+pub struct PedersenHash {
+    state: Felt252,
+}
+
+impl PedersenHash {
+    pub fn init_digest_to(digest: Vec<u8>) -> Self {
```
rot256: This could fail: if the digest is too long this should result in an Error or Option::None

rot256 (approved):
Overall looks good.

Mostly nitpicking on my end.

pedersen/src/pedersen.rs
varunthakore: I am using `BigUint` (https://docs.rs/num-bigint/latest/num_bigint/struct.BigUint.html) and not `Bigint<N>` because `BigUint` has functions for `/` and `%`.

pedersen/src/pedersen.rs
varunthakore: I have used the following as reference where `FieldElementT = PrimeFieldElement<252, 0>` i.e. base field of the curve.
https://github.com/starkware-libs/stone-prover/blob/1414a545e4fb38a85391289abe91dd4467d268e1/src/starkware/crypt_tools/pedersen.inl#L40-L64

pedersen/src/pedersen.rs
varunthakore: Here also, I have used the following as reference:
https://github.com/starkware-libs/stone-prover/blob/1414a545e4fb38a85391289abe91dd4467d268e1/src/starkware/crypt_tools/pedersen.inl#L40-L64

PR #26, opened 2024-10-21
Implement Composition Polynomial

varunthakore:
This PR implements the composition polynomial library.
https://github.com/starkware-libs/stone-prover/tree/main/src/starkware/composition_polynomial

### DIFF
--- a/composition_polynomial/src/air.rs
+++ b/composition_polynomial/src/air.rs
@@ -1,0 +1,218 @@
+use crate::composition_polynomial::{batch_pow, CompositionPolynomial};
+use crate::periodic_columns::PeriodicColumn;
+use ark_ff::PrimeField;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+pub trait Air<F: PrimeField> {
+    fn create_composition_polynomial(
+        &self,
+        trace_generator: &F,
+        random_coefficients: &[F],
+    ) -> CompositionPolynomial<F>;
+
+    fn trace_length(&self) -> usize;
+
+    fn get_composition_polynomial_degree_bound(&self) -> usize;
+
+    fn num_random_coefficients(&self) -> usize;
+
+    fn get_num_constraints(&self) -> usize {
+        self.num_random_coefficients()
+    }
+
+    fn constraints_eval(
+        &self,
+        neighbors: &[F],
+        periodic_columns: &[F],
+        random_coefficients: &[F],
+        point: &F,
+        gen_powers: &[F],
+        precomp_domains: &[F],
+    ) -> F;
+
+    fn get_mask(&self) -> &[(usize, usize)];
+
+    fn num_columns(&self) -> usize;
+
+    fn domain_evals_at_point(&self, point_powers: &[F], shifts: &[F]) -> Vec<F>;
+
+    fn parse_dynamic_params(&self, params: &HashMap<String, usize>) -> Vec<usize>;
+
+    fn with_interaction_elements(&self, _interaction_elms: &[F]) -> Box<dyn Air<F>> {
+        panic!("Calling with_interaction_elements in an air with no interaction.");
+    }
+
+    fn get_interaction_params(&self) -> Option<InteractionParams>;
+
+    fn get_n_columns_first(&self) -> usize {
+        match self.get_interaction_params() {
+            Some(params) => params.n_columns_first,
+            None => self.num_columns(),
+        }
+    }
+
+    fn precompute_domain_evals_on_coset(
+        &self,
+        point: &F,
+        generator: &F,
+        point_exponents: &[usize],
+        shifts: &[F],
+    ) -> Vec<Vec<F>>;
+}
+
+pub struct InteractionParams {
+    pub n_columns_first: usize,
+    pub n_columns_second: usize,
+    pub n_interaction_elements: usize,
+}
+
+type ConstraintFunction<F> = Arc<dyn Fn(&[F], &[F], &[F], &F, &[F], &[F]) -> F>;
+
+// DummyAir struct definition
+#[derive(Clone)]
+pub struct DummyAir<F: PrimeField> {
+    pub trace_length: usize,
+    pub n_constraints: usize,
+    pub n_columns: usize,
+    pub mask: Vec<(usize, usize)>,
+    pub periodic_columns: Vec<PeriodicColumn<F>>,
+    pub point_exponents: Vec<usize>,
+    pub gen_exponents: Vec<usize>,
+    pub constraints: Vec<ConstraintFunction<F>>,
+    pub composition_polynomial_degree_bound: Option<usize>,
+}
+
+impl<F: PrimeField> DummyAir<F> {
+    pub fn new(trace_length: usize) -> Self {
+        assert!(trace_length.is_power_of_two());
+        Self {
+            trace_length,
+            n_constraints: 0,
+            n_columns: 0,
+            mask: vec![],
+            periodic_columns: vec![],
+            point_exponents: vec![],
+            gen_exponents: vec![],
+            constraints: vec![],
+            composition_polynomial_degree_bound: None,
+        }
+    }
+}
+
+impl<F: PrimeField> Air<F> for DummyAir<F> {
+    fn trace_length(&self) -> usize {
+        self.trace_length
+    }
+
+    fn get_composition_polynomial_degree_bound(&self) -> usize {
+        assert!(
+            self.composition_polynomial_degree_bound.is_some(),
+            "composition_polynomial_degree_bound wasn't initialized."
+        );
+        self.composition_polynomial_degree_bound.unwrap()
+    }
+
+    fn num_random_coefficients(&self) -> usize {
+        self.n_constraints
+    }
+
+    fn num_columns(&self) -> usize {
+        self.n_columns
+    }
+
+    fn get_interaction_params(&self) -> Option<InteractionParams> {
+        None
+    }
+
+    fn constraints_eval(
+        &self,
+        neighbors: &[F],
+        periodic_columns: &[F],
+        random_coefficients: &[F],
+        point: &F,
+        gen_powers: &[F],
+        precomp_domains: &[F],
+    ) -> F {
+        assert!(
+            random_coefficients.len() == self.constraints.len(),
+            "This is a bug in the test."
+        );
+
+        let mut res = F::ZERO;
+        for constraint in self.constraints.iter() {
+            res += constraint(
+                neighbors,
+                periodic_columns,
+                random_coefficients,
+                point,
+                gen_powers,
+                precomp_domains,
+            );
+        }
+        res
+    }
+
+    fn domain_evals_at_point(&self, point_powers: &[F], _shifts: &[F]) -> Vec<F> {
+        if point_powers.len() <= 1 {
+            return vec![];
+        }
+        vec![point_powers[1] - F::ONE]
+    }
+
+    fn create_composition_polynomial(
+        &self,
+        trace_generator: &F,
+        random_coefficients: &[F],
+    ) -> CompositionPolynomial<F> {
+        let shifts = batch_pow(trace_generator, &self.gen_exponents);
+        CompositionPolynomial::new(
+            Box::new(self.clone()),
+            *trace_generator,
+            self.trace_length(),
+            self.periodic_columns.clone(),
+            random_coefficients,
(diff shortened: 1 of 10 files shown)


composition_polynomial/src/air.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub n_interaction_elements: usize,
+}
+
+type ConstraintFunction<F> = Arc<dyn Fn(&[F], &[F], &[F], &F, &[F], &[F]) -> F>;
+
+// DummyAir struct definition
+#[derive(Clone)]
+pub struct DummyAir<F: PrimeField> {
+    pub trace_length: usize,
+    pub n_constraints: usize,
```
rot256: Is this different from `self.constraints.len()`?

composition_polynomial/src/air.rs
```diff
@@ -0,0 +1,218 @@
...
+    ) -> Vec<Vec<F>>;
+}
+
+pub struct InteractionParams {
+    pub n_columns_first: usize,
+    pub n_columns_second: usize,
+    pub n_interaction_elements: usize,
+}
+
+type ConstraintFunction<F> = Arc<dyn Fn(&[F], &[F], &[F], &F, &[F], &[F]) -> F>;
```
rot256: Consider introducing a struct for the argument: that way they are named and it is easy to see what a `ConstraintFunction` takes e.g.

```rust
struct ConstraintEval {
  neighbors: &[F]
  periodic_columns: &[F]
  random_coefficients: $[F]
  point: &[F]
  gen_powers: &[F]
  precomp_domains: &[F]
}
```

composition_polynomial/src/air.rs
```diff
@@ -0,0 +1,218 @@
...
+        precomp_domains: &[F],
+    ) -> F {
+        assert!(
+            random_coefficients.len() == self.constraints.len(),
+            "This is a bug in the test."
+        );
+
+        let mut res = F::ZERO;
+        for constraint in self.constraints.iter() {
+            res += constraint(
```
rot256: I thought the `random_coefficients` where used to combine each constraint? Does each constraint need to take the random coefficients or can we compute the combination here?

composition_polynomial/src/air.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    fn num_columns(&self) -> usize {
+        self.n_columns
+    }
+
+    fn get_interaction_params(&self) -> Option<InteractionParams> {
+        None
+    }
+
+    fn constraints_eval(
```
rot256, line 155: Could this just be a method on `CompositionPolynomial<F>`?

e.g. it takes `random_coefficients` which is really what defines a `CompositionPolynomial` from an AIR. So if we move it there we would not need to take e.g. `random_coefficients`

composition_polynomial/src/air.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    fn domain_evals_at_point(&self, point_powers: &[F], _shifts: &[F]) -> Vec<F> {
+        if point_powers.len() <= 1 {
+            return vec![];
+        }
+        vec![point_powers[1] - F::ONE]
+    }
+
+    fn create_composition_polynomial(
```
rot256, line 190: Could this take ownership of the Air?

i.e. 

```rust
fn into_composition_polynomial(
   self,
   trace_generator,
   random_coefficients
) -> CompositionPolynomial<F>
```

That way we avoid the Box and Clone?

composition_polynomial/src/composition_polynomial.rs
```diff
@@ -0,0 +1,295 @@
...
+            neighbors,
+            &periodic_column_vals,
+            &self.coefficients,
+            point,
+            &self.shifts,
+            &domain_evals,
+        )
+    }
+
+    pub fn eval_on_coset_bit_reversed_output(
```
rot256, line 74: Whats the relation between `eval_on_coset_bit_reversed_output` and `eval_at_point`?

composition_polynomial/src/composition_polynomial.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        for point_idx in 0..self.coset_size {
+            out_evaluation[bit_reverse(point_idx, log_coset_size)] = batch_inverse[point_idx];
+        }
+
+        out_evaluation
+    }
+}
+
+pub fn batch_pow<F: PrimeField>(base: &F, exponents: &[usize]) -> Vec<F> {
```
rot256, line 145: This is fine. 

But FYI, there are more efficient ways of doing this using [addition chains](https://en.wikipedia.org/wiki/Addition_chain).

composition_polynomial/src/composition_polynomial.rs
```diff
@@ -0,0 +1,295 @@
...
+    let mut output = Vec::with_capacity(exponents.len() + 1);
+    output.push(*base);
+
+    for e in exponents {
+        output.push(base.pow([*e as u64]));
+    }
+    output
+}
+
+fn bit_reverse(n: usize, number_of_bits: usize) -> usize {
```
rot256, line 155: https://doc.rust-lang.org/std/primitive.u64.html#method.reverse_bits

Combined with a single shift:


```rust
(n as u64).reverse_bits() >> (64 - number_of_bits)
```

rot256: Maybe add an assert as well:

```rust
debug_assert!(n < (1 << number_of_bits))
```

composition_polynomial/src/composition_polynomial.rs
```diff
@@ -0,0 +1,295 @@
...
+    use ark_ff::{Field, PrimeField, UniformRand};
+    use felt::Felt252;
+    use rand::Rng;
+    use std::sync::Arc;
+
+    fn get_subgroup_generator<F: PrimeField>(n: usize) -> F {
+        let q_minus_1: num_bigint::BigUint = F::ONE.neg().into();
+
+        // Calculate (q - 1) / n
+        assert!(
```
rot256: nitpick: assert_eq

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn new(
+        values: Vec<F>,
+        group_generator: F,
+        offset: F,
+        coset_size: usize,
+        column_step: usize,
+    ) -> Self {
+        let period_in_trace = values.len() * column_step;
+
+        assert!(period_in_trace != 0);
```
rot256: nit: assert_ne

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+        assert!(coset_size % period_in_trace == 0);
```
rot256: nit: assert_eq

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+    period_in_trace: usize,
+    n_copies: usize,
+    offset_compensation: F,
+    lde_manager: MultiplicativeLDE<F>,
+}
+
+impl<F: PrimeField> PeriodicColumn<F> {
+    pub fn new(
+        values: Vec<F>,
+        group_generator: F,
```
rot256, line 18: Since we are "passing around" subgroups and cosets, it might actually make sense to abstract this into a separate type, e.g.

```rust
struct Group<F> {
   gen: F,
   order: usize
}
```

```rust
struct Coset<F> {
   group: Group<F>,
   shift: F
}
```

With appropriate methods on them?

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+        offset: F,
+        coset_size: usize,
```
rot256, line 20: Would get rid of these arguments as well.

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    pub fn get_actual_degree(&self) -> usize {
+        self.lde_manager.ldes[0].degree()
+    }
+
+    pub fn get_coset(&self, start_point: &F, coset_size: usize) -> CosetEvaluation<F> {
+        let offset = start_point.pow([self.n_copies as u64]);
+        let n_values = self.lde_manager.base.size as usize;
+
+        assert!(
```
rot256: nit: assert_eq

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+pub struct CosetEvaluation<F: PrimeField> {
+    index_mask: usize,
+    values: Vec<F>,
+}
+
+impl<F: PrimeField> CosetEvaluation<F> {
+    pub fn new(values: Vec<F>) -> Self {
```
rot256, line 92:
```rust
assert!(values.is_power_of_two())
```

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+        coset_size: usize,
+        column_step: usize,
+    ) -> Self {
+        let period_in_trace = values.len() * column_step;
+
+        assert!(period_in_trace != 0);
+        assert!(coset_size % period_in_trace == 0);
+        let n_copies = coset_size / period_in_trace;
+
+        let offset_compensation = offset.pow([n_copies as u64]).inverse().unwrap();
```
rot256, line 29: What's the difference between `start_point` (as used in `get_coset`) and `offset`? Both should be the first evaluation point of the first entry of the first copy of the periodic table?

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+
+impl<F: PrimeField> PeriodicColumn<F> {
+    pub fn new(
+        values: Vec<F>,
+        group_generator: F,
+        offset: F,
+        coset_size: usize,
+        column_step: usize,
+    ) -> Self {
+        let period_in_trace = values.len() * column_step;
```
rot256, line 23: All these are restricted to powers of two?

composition_polynomial/src/periodic_columns.rs
```diff
@@ -0,0 +1,199 @@
...
+            group_generator,
+            column_step,
+            period_in_trace,
+            n_copies,
+            offset_compensation,
+            lde_manager,
+        }
+    }
+
+    pub fn eval_at_point(&self, x: F) -> F {
```
rot256, line 45: Nvm: the LDE manager is used to interpolate exactly this polynomial.

~Why do we need the `lde_manager`?~

We have `values` and assuming the length of the table divides the group order the polynomial for the column has a succinct description (and can be evaluated efficiently). So can't we just evaluate it directly?

For instance, the polynomial f(X) which evaluates to:

```
f(1)   = y0
f(w)   = y1
f(w^2) = y0
f(w^3) = y1
...
```

i.e. alternating. For a domain of length $2^k$, is simply:

$$
f(X) = a + b * X^{2^k / 2}
$$

Where $a, b$: 

$$
f(1) = a + b = y0
$$

$$
f(w) = a + b * w^{2^k / 2} = a - b = y1
$$

So $a = (y0 + y1) / 2$ and $b = (y0 - y1) / 2$. A similar deal applies when you have a "starting_point", except now the polynomial is composed with * starting_point.


Here is a Sage script if you want to play around this:

```sage
F = GF(2^251 + 17 * 2^192 + 1)
k = 192

g = F.multiplicative_generator()
g = g^(g.multiplicative_order() / 2^k)

order = 2^8
w = g^(g.multiplicative_order() / order)

print(f"order: {order}")
print(g.multiplicative_order())

table = [
    1, 2, 1, 2
]

ys = table * (order // len(table))
xs = [w^i for i in range(order)]

R.<X> = PolynomialRing(F)

f = R.lagrange_polynomial(zip(xs, ys))

print(f)
```

Otherwise, let's hop on a call.

composition_polynomial/src/air.rs
varunthakore: its the same, removed `n_constraints`.

composition_polynomial/src/composition_polynomial.rs
varunthakore: If you evaluate the polynomial on points `coset_offset*<group_genertor>` using `eval_at_point`, these evaluations should be same as the bit reversed output of `eval_on_coset_bit_reversed_output`.

PR #27, opened 2024-10-28
Implement Boundary Constraints

varunthakore:
This PR implements boundary constraints which are part of the following library.
https://github.com/starkware-libs/stone-prover/tree/main/src/starkware/air

This PR has dependency on #26.

### DIFF
--- a/air/src/boundry.rs
+++ b/air/src/boundry.rs
@@ -1,0 +1,271 @@
+use ark_ff::PrimeField;
+use composition_polynomial::{air::Air, composition_polynomial::CompositionPolynomial};
+
+#[derive(Clone)]
+pub struct BoundaryAir<F: PrimeField> {
+    trace_length: usize,
+    n_columns: usize,
+    constraints: Vec<ConstraintData<F>>,
+    mask: Vec<(usize, usize)>,
+}
+
+#[derive(Clone)]
+pub struct ConstraintData<F> {
+    coeff_idx: usize,
+    column_index: usize,
+    point_x: F,
+    point_y: F,
+}
+
+impl<F: PrimeField> BoundaryAir<F> {
+    pub fn new(
+        trace_length: usize,
+        n_columns: usize,
+        boundary_conditions: &[(usize, F, F)],
+    ) -> Self {
+        let mut constraints = Vec::with_capacity(boundary_conditions.len());
+
+        for (coeff_idx, &(column_index, ref point_x, ref point_y)) in
+            boundary_conditions.iter().enumerate()
+        {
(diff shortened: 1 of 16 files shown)


air/src/boundry.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut constraints = Vec::with_capacity(boundary_conditions.len());
+
+        for (coeff_idx, &(column_index, ref point_x, ref point_y)) in
+            boundary_conditions.iter().enumerate()
+        {
+            let x = *point_x;
+
+            let pos = constraints
+                .iter()
+                .position(|constraint: &ConstraintData<F>| constraint.point_x == x);
```
rot256, line 35: Might be fine since the number of boundary constraints is probably small, but this has O(n^2) complexity.

Maybe use a Map : point_x -> point_y

Then collect into ConstraintData at the end.

rot256: What is the use case of "overwriting" a previous boundary constraint with a different point_y? Should we just error instead?

air/src/boundry.rs
```diff
@@ -0,0 +1,271 @@
...
+                constraints.insert(pos, constraint_data);
+            } else {
+                constraints.push(constraint_data);
+            }
+        }
+
+        let mut mask = Vec::with_capacity(n_columns);
+        for i in 0..n_columns {
+            mask.push((0, i));
+        }
```
rot256, line 54: `let mask = (0..n_columns).map(|i| (0, i)).collect()`

air/src/boundry.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    fn constraints_eval(
+        &self,
+        neighbors: &[F],
+        _periodic_columns: &[F],
+        random_coefficients: &[F],
+        point: &F,
+        _shifts: &[F],
+        _precomp_domains: &[F],
+    ) -> F {
```
rot256, line 90: How do we compose this with other types of constraints?

rot256 (commented):
The only delta compared to the other PR is `air/src/boundary.rs` right?

varunthakore:
> The only delta compared to the other PR is `air/src/boundary.rs` right?

The code in the following commit is the delta
https://github.com/zksecurity/pumice/pull/27/commits/c57fc90cd4e7f3766e58a092cac537e32a68e2de

air/src/boundry.rs
varunthakore: I haven't reviewed the Stone code where boundary constraints are composed with other types of constraints, so I can't answer this yet.

